PR #1, opened 2019-10-25
PTI-68: Initial Cleanup

NickPhura:
This is probably too gigantic to review, as its the initial commit of the entire code base, but here is a rough list of changes made:

- Removed unused code, including code that may eventually be used, but would currently be broken.

- Removed unused package.json dependencies.
- Bumped all pacakge.json versions.

- Ensured existing admin unit tests pass.
- Most api unit tests also pass, but more work is still needed because the api is fairly gutted at the moment.

- Updated old references to ACRFD to NRPTI instead, etc.

- All Linters pass.
- All code that was specifically edited/modified should have JSDoc (very work in progress code, and old code, may not).

- Work on PTI-80 and PTI-81 is also in here, but it is still WIP.


NickPhura:
Just throwing everyone on these reviews, and whoever has the time to review them can do so.

Obviously for this one its probably not reviewable, but in future.

PR #10, opened 2019-10-31
PTI-80: Update keycloak connection strings.

NickPhura:
- Updated some linting issues.
- Updated a few package versions to satisfy some of the dependency Bot issues.
- Updated .gitignore to remove lock files.
- Update mongoose settings to address deprecation warning.


admin/package.json
```diff
@@ -44,7 +44,7 @@
     "keycloak-angular": "5.1.0",
     "leaflet": "1.5.1",
     "linkify-it": "2.2.0",
-    "lodash": "4.17.15",
+    "lodash": "4.17.13",
```
danieltruong: Remove lodash

public-lng/package.json
```diff
@@ -45,7 +45,7 @@
     "leaflet.markercluster": "1.4.1",
     "leaflet": "1.4.0",
     "linkify-it": "2.2.0",
-    "lodash": "4.17.15",
+    "lodash": "4.17.13",
```
danieltruong: Remove

admin/package.json
NickPhura: This might be better as its own jira task, since lodash is used in the code in about 100 places currently.

NickPhura: Created a ticket: https://bcmines.atlassian.net/browse/PTI-140

PR #17, opened 2019-11-14
Removing lodash from api

danieltruong:


api/api/utils/query-utils.js
```diff
@@ -18,8 +17,8 @@ let DEFAULT_PAGESIZE = 100;
  * @returns array of fields that is a subset of allowedFields.
  */
 exports.getSanitizedFields = function(allowedFields, fields) {
-  return _.remove(fields, function(field) {
-    return _.indexOf(allowedFields, field) !== -1;
+  return fields.filter(function(field) {
+    allowedFields.indexOf(allowedFields, field) !== -1;
```
NickPhura: I think this needs a 'return'.

return allowedFields..

PR #27, opened 2019-11-22
PTI-167: Added store service. Added breadcrumb functionality.

danieltruong:


angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  set state(state: any) {
+    this[state.type] = state.data;
+  }
+
+  public getNavigationStack() {
+    if (this.state.navigationStack) {
+      return this.state.navigationStack;
+    } else {
+      return null;
+    }
```
NickPhura: Do we need to check if `this.state` is non-null before checking `this.state.navigationStack`? Unless `this.state` can never be null.

(Optional) you could write it like this, as an alternate pattern to if/else.  But its the same output in the end in either case:
`return this.state && this.state.navigationStack || null`

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  constructor() {
+    this.currentState = {};
+    this.state.navigationStack = [];
+  }
+
+  get state(): any {
+    return this.currentState;
+  }
+  set state(state: any) {
```
NickPhura: Might be a good idea to replace `any` with some kind of state type, so consumers know to give it a `type` and `data` property.

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+
+  public getLastBackUrl() {
+    const stack = this.getNavigationStack();
+    if (stack) {
+      return stack[stack.length - 1].backUrl;
+    } else {
+      return null;
+    }
```
NickPhura: Similar to above.  We probably need to check that `stack[stack-length-1]` is non-null, before trying to reference `backUrl` from it.

(Optional) Same alternate pattern as above, this could be written as:
`return stack && stack[stack.length - 1] && stack[stack.length - 1].backUrl || null`

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public pushNavigationStack(backUrl: string[], breadcrumbs: object[], componentId: string = null) {
+    const navigationObject = {
+      backUrl: backUrl,
+      breadcrumbs: breadcrumbs,
+      componentId: componentId
+    };
```
NickPhura: It would be good to turn this into a type (interface), and apply it anywhere its relevant.

Then additionally, instead of expecting 3 parameters in this method, you could just take in a one parameter. 
Something like:
`public pushNavigationStack(navigaionObject: INavigationObject) {`

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+        this.state.navigationStack = stack;
+      }
+      return stackObject;
+    } else {
+      this.clearNavigationStack();
+      return null;
+    }
+  }
+
+  public navigateBreadcrumb(breadcrumb, router) {
```
NickPhura: Should give these 2 params types

angular/projects/global/src/lib/services/store.service.ts
NickPhura: Just read further down and see that there already is a navigation type.  So maybe just needs to be used here where relevant.

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+/*
+  Example of a back Url array:
+  ['/p', this.project._id, 'edit']
+
+  Example of a breadcrumbs array:
+  Note that the objects in the array contain a route and a label.
+  [
+      {
+          route: ['/projects'],
+          label: 'All Projects'
```
NickPhura: You could create a breadcrumbs type to use, instead of `object`.  Then you can enforce `route` and `label` as properties.

`breadcrumbs: IBreadcrumb[]`

angular/projects/global/src/lib/services/store.service.ts
```diff
@@ -0,0 +1,153 @@
...
+      throw Error('You must pass in a breadcrumbs');
+    }
+
+    this.breadcrumbs = (params && params.breadcrumbs) || [];
+    this.componentName = (params && params.componentName) || '';
+    this.componentId = (params && params.componentId) || '';
+
+    // Set backRef
+    if (this.breadcrumbs.length > 1) {
+      this.backRef = this.breadcrumbs[this.breadcrumbs.length - 2].route;
```
NickPhura: Probably I just haven't wrapped my head around this enough, but wouldn't the backref just be 1 element back, and not 2?

NickPhura: Nevermind I get it.  Because `breadcrumbs.length` is 1 bigger than the last index, so -2 is just the previous index.

angular/projects/global/src/lib/services/store.service.ts
danieltruong: State can literally be anything. So you are able to set it to an array, form, string, custom object, etc.

PR #37, opened 2019-11-26
PTI-167: Added table template utils.

danieltruong:


angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+  DEFAULT_TABLE_CURRENT_PAGE,
+  DEFAULT_TABLE_SORT_BY,
+  DEFAULT_TABLE_KEYWORDS
+} from './table-object';
+
+@Injectable()
+export class TableTemplateUtils {
+  constructor(private platformLocation: PlatformLocation, private router: Router) {}
+
+  public updateUrl(tableObject: TableObject) {
```
NickPhura: This function is fairly large, and does a lot of stuff.

It would be my recommendation to break it up into a bunch of small functions, each of which is given just the piece of tableObject they care about, and can handle their own error handling and return values, etc.  And then this function is reduced to just managing currentUrl, and calling all of the other functions, and updating currentUrl based on their responses.  

You then can document each of the smaller functions more easily, without dumping tons of inline comments into this function.  And each of them is now testable in isolation, and your tests will become smaller as a result.   Making updates becomes easier as well, since each function handles all of its own input and output, and doesn't care about what may have come before or after it.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+          currentUrl += `;${key}=`;
+          tableObject.filter[key].split(',').forEach(item => {
+            currentUrl += `${item},`;
+          });
+          currentUrl = currentUrl.slice(0, -1);
+        }
+      });
+    }
+    currentUrl += ';ms=' + new Date().getTime();
```
NickPhura: I've seen this pattern before.  As far as I know it shouldn't be necessary? As long as we subscribe to the route and update the page accordingly when params change.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+    filterFieldList.map(field => {
+      if (routeParams[field]) {
+        tableObject.filter[field] = routeParams[field];
+      }
+    });
+
+    // TODO: Get fields from URL.
+
+    if (forceUpdateUrl) {
+      this.updateUrl(tableObject);
```
NickPhura: This looks like its purely for convenience, so someone can update the table object and the url in one call, rather than having to call both.

This is just my preference, but I like it when this sort of thing doesnt exist.  If the consumer makes 2 calls (updateUrl(), updateTableObject()) its really clear what that does.  But if you just see (updateTableObjct(true)) its not as clear immediately what that is doing.  

It also becomes harder to make updates to these functions, because you need to track down who is calling it with true, and who with false, in order to know what other code you might affect.  Whereas if this function only does 1 thing, then its obvious what the affect will be.

It also technically makes your testing slightly more complex, because now you need to test the path where forceUpdateUrl is true and false, on top of everything else this function does.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,87 @@
...
+    return tableObject;
+  }
+
+  public navigateUsingParams(navigationObject: TableObject, path: any[]) {
+    const params = this.getNavParamsObj(navigationObject);
+    path.push(params);
+    this.router.navigate(path);
+  }
+
+  public getNavParamsObj(navigationObject: TableObject) {
```
NickPhura: General comment for any function.  Make sure to check if params are null, before accessing their values, otherwise we will get exceptions.  It also means we can exit the function immediately which is more efficient (for you optimization guys).

danieltruong: The required param is a nav object. Upon creating a nav object, all fields will be filled with a default.

PR #40, opened 2019-11-27
PTI-167: Added table template utils.

danieltruong:


angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  public getBaseUrl(): string {
+    let currentUrl = this.router.url;
+    currentUrl = (this.platformLocation as any).getBaseHrefFromDOM() + currentUrl.slice(1);
+    currentUrl = currentUrl.split(';')[0];
+    return currentUrl;
+  }
+
+  public setPaginationInUrl(currentUrl: string, currentPage: number, pageSize: number) {
+    currentUrl += `;currentPage=${currentPage ? currentPage : DEFAULT_TABLE_CURRENT_PAGE};pageSize=${
```
NickPhura: Instead of using a ternary:``` currentPage ? currentPage : DEFAULT_TABLE_CURRENT_PAGE ```
I would use the short circuit style: ``` currentPage || DEFAULT_TABLE_CURRENT_PAGE ```

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+      pageSize ? pageSize : DEFAULT_TABLE_PAGE_SIZE
+    }`;
+    return currentUrl;
+  }
+
+  public setKeywordsInUrl(currentUrl, keywords) {
+    if (keywords !== null && keywords !== '') {
```
NickPhura: Instead of: ``` keywords !== null && keywords !== '' ```
It should be sufficient to just say: ``` !keywords ```
Since null and empty string are both falsey values.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  public setSortByInUrl(currentUrl, sortBy) {
+    if (sortBy !== null && sortBy !== '') {
+      currentUrl += `;sortBy=${sortBy}`;
+    }
+    return currentUrl;
+  }
+
+  public setFilterInUrl(currentUrl, filter) {
+    if (filter !== null && filter !== {}) {
```
NickPhura: In this case, Id use: ``` !filter && filter !== {} ```

The other benefit of !<variable> is that it also covers the case where its undefined.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return currentUrl;
+  }
+
+  public updateTableObjectWithUrlParams(routeParams: Params, tableObject: TableObject, filterFieldList = []) {
+    tableObject.pageSize = routeParams.pageSize ? +routeParams.pageSize : DEFAULT_TABLE_PAGE_SIZE;
+    // tslint:disable-next-line: max-line-length
+    tableObject.currentPage = routeParams.currentPage ? +routeParams.currentPage : DEFAULT_TABLE_CURRENT_PAGE;
+    tableObject.sortBy = routeParams.sortBy ? routeParams.sortBy : DEFAULT_TABLE_SORT_BY;
+    tableObject.keywords = routeParams.keywords ? routeParams.keywords : DEFAULT_TABLE_KEYWORDS;
```
NickPhura: Could replace with the short circuit style.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    });
+
+    // TODO: Get fields from URL.
+
+    return tableObject;
+  }
+
+  public navigateUsingParams(tableObject: TableObject, path: any[]) {
+    if (tableObject == null) {
```
NickPhura: Recommend using ``` !tableOjbect ```
Or at least a triple === if we specifically want to check for null.

angular/projects/global/src/lib/components/table-template/table-template-utils.ts
```diff
@@ -0,0 +1,117 @@
...
+      throw Error('Navigation Object cannot be null.');
+    }
+
+    if (path == null || path.length === 0) {
```
NickPhura: Similar to above, you could just say ``` !path || !path.length ```

NickPhura:
Also, I forgot to add: make sure to throw types on all of the parameters

PR #56, opened 2019-12-13
PTI-264: Epic Data Connection + Integration with Tasks.

NickPhura:
Things of note, in no particular order:

It currently supports running 1 datasource (Epic) for 1 of the supported types (Order or Inspection).

I have written some tests, but not all yet.

This breaks the record list page on admin-nrpti, since the api now uses 'Orders' and 'Inspections', but the frontend still expects 'Records'

It currently hardcodes a milestone id (whichever Daniel had been using before).  I need to get a better understanding for what the milestone is for, and how to fetch the one we want, etc.

I put a TODO placeholder in epic-datasource.js for fetching/streaming the documents to S3.  It wouldn't necessarily have to go in that exact location, but I think somewhere around there?  Might depend on how documents from Epic are fetched, and what the S3 integration looks like.


api/src/integrations/epic/epic-datasource.js
```diff
@@ -0,0 +1,191 @@
...
+
+      status.epicStatus = 'no epic records found';
+      return status;
+    }
+
+    // Update task record with number of epic records found
+    this.taskAuditRecord.updateTaskRecord({ itemTotal: epicRecords.length });
+
+    // Add epic meta to status (optional)
+    const epicMeta = data && data[0] && data[0].meta && data[0].meta[0];
```
marklise: `const epicMeta = data && data[0] && data[0].meta && data[0].meta[0];`
to
`const epicMeta = data && data[0] && data[0].meta && data[0].meta.length;`

api/src/tasks/import-task.js
```diff
@@ -1,169 +1,68 @@
...
-  // Job has come in
-  let scopes = args.swagger.params.auth_payload.realm_access.roles;
+  const scopes = args.swagger.params.auth_payload.realm_access.roles;
   defaultLog.info('scopes:', scopes);
 
-  // Inspect payload
+  const taskAuditRecord = new TaskAuditRecord();
+  taskAuditRecord.updateTaskRecord({
+    dataSourceLabel: args.swagger.params.task.value.dataSource,
+    startDate: new Date()
```
marklise: See latest changes to master - needs read/write props.

angular/projects/admin-nrpti/src/app/import/import.component.ts
```diff
@@ -119,6 +119,6 @@ export class ImportComponent implements OnInit {
   }
 
   postToApi(): Observable<any> {
-    return this.factoryService.startTask({ dataSource: 'epic' });
+    return this.factoryService.startTask({ dataSource: 'epic', recordType: 'inspection' });
```
NickPhura: Added this here for now, as currently the import-task expects/requires exactly 1 dataSource and 1 recordType.

api/openshift/jobs/integrations.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: batch/v1beta1
```
NickPhura: This is outdated.  

Could remove it even now that we aren't focusing on scheduled jobs anymore.

NickPhura:
Closed as I need to add more functionality after checking ticket.

PR #65, opened 2019-12-17
PTI-264: Epic Data Connection + Tasks Integration + Unit Tests.

NickPhura:


api/src/models/index.js
```diff
@@ -1,5 +1,17 @@
...
+// general
+exports.audit = require('./audit');
+exports.task = require('./task');
+
+// master
+exports.order = require('./master/order');
+exports.inspection = require('./master/inspection');
+
+// lng
+exports.orderLNG = require('./lng/order-lng');
```
marklise: This should be importing the directory, by specifying an index.js, instead of the specific items in this index.js (same for next few lines)

api/src/integrations/epic/epic-orders.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param {Order} orderRecord NRPTI Order record (required)
+   * @returns {string} status of the add/update operations.
+   * @memberof EpicOrders
+   */
+  async saveRecord(orderRecord) {
+    if (!orderRecord) {
+      throw Error('saveRecord - required record must be non-null.');
+    }
+
+    // TODO Currently this always adds new records. Need to support updating existing records?
```
marklise: Correct - we should not be inserting dupes.

NickPhura: Sorry, that comment is actually out of date and should be removed.

That was the case last week, but I've since changed it to use findOneAndUpdate  + upsert to not add duplicates.

api/src/models/index.js
NickPhura: Fixed.

api/src/integrations/epic/epic-orders.js
NickPhura: Fixed (removed comment).

api/src/integrations/epic/epic-orders.js
```diff
@@ -0,0 +1,102 @@
...
+  async saveRecord(orderRecord) {
+    if (!orderRecord) {
+      throw Error('saveRecord - required record must be non-null.');
+    }
+
+    try {
+      const Order = mongoose.model('Order');
+
+      const record = await Order.findOneAndUpdate(
+        { documentId: orderRecord.documentId },
```
NickPhura: @marklise This is currently how it determines updates vs adds (same for epic-inspections.js).

PR #67, opened 2019-12-21
PTI-81: Update list page. Add page size control. Add page count display.

NickPhura:
Tweaked some of the general UI, so the header bars and side nav sync up tighter.

Added 2 components to the table template: page-size picker and a page-count display.

Updated all of the models to what I believe (at this point in time) is the latest version.

Made all currently existing *.spec.ts files pass, so we at least have a clean slate to work from (with whatever amount of testing we end up adding/not adding).


api/src/controllers/search.js
```diff
@@ -4,6 +4,7 @@ var mongoose = require('mongoose');
 var Actions = require('../utils/actions');
```
NickPhura: Apparently you cant comment on lines that haven't changed.  So this is for line 2 above.

For logging, we can use the logger.js class I set up.  Adds some extra formatting and prepends the name you pass in to the logs so they read nicer.

```
const defaultLog = require('../utils/logger')('search');
```

Not an issue for this PR, but just in general for future stuff.

PR #68, opened 2019-12-21
PTI-81-2: Add .toArray() call to mongodb aggregate cursor result.

NickPhura:
Add more error handling to list pages for when results are empty.

Was checking out the new mongo stuff, but noticed we were getting a circular json stringify error.  Investigated and it looks like the mongo aggregate by default returns the entire cursor.  But slapping on .toArray() gives just the results.

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -325,7 +325,7 @@ var searchCollection = async function (roles, keywords, schemaName, pageNum, pag
         defaultLog.info("err:", err);
         resolve([]);
       } else {
-        resolve(data);
+        resolve(data.toArray());
       }
     })
   });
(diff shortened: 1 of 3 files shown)


api/src/controllers/search.js
```diff
@@ -325,7 +325,7 @@ var searchCollection = async function (roles, keywords, schemaName, pageNum, pag
         defaultLog.info("err:", err);
         resolve([]);
       } else {
-        resolve(data);
+        resolve(data.toArray());
```
NickPhura: The only thing I didnt look into fully was if resolve(new Promise()) is okay or not (.toArray() returns a promsie).  I know the entire function is being awaited, but not 100% sure if this is fine still.

I did some basic tests just in the web console and it seemed fine...

marklise: This is ok

PR #85, opened 2020-01-07
PTI-126/PTI-4: Public NRCED list + order details pages + table template updates

NickPhura:
- Update Table-templates to inject row components
  - It still accepts a top level component parameter, if you know you only ever want 1 type of row component.  Individual items components will override this on a row-by-row basis, if set.
  - supports a messageIn and messageOut event emitter, so child rows can emit to all parent levels, and the parents can emit to all child rows.  To trigger whatever events, as needed in either direction.

- Add public card list
  - In regards to table-template, and the public card toggling: There are 2 ways this can be done now.
     1. Let your row component handle toggling its own state, and then leverage the messageIn/Out to tell the parent to tell all children to collapse, or w/e effect you want. (which is what is currently set up)
     2. Update the TableObject.items array to change the component.  This will trigger the table to re-render the new component.  Caveat: the component you render needs to know its rendering as a row, so it would need to have a tr[...] selector, etc.

- Add initial Order Details page
  - Nothing special here yet, just super basic placeholder UI

- Add batch records publish/unpublish
  - records/publish and records/unpublish
  - they take a body parameter that contains an array of {recordId, recordType} objects.

- Various misc tweaks

- Update all uses of table template
  - All row components now extend a base TableRowComponent class, which contains the data inputs and messageIn/Out emitters. So you dont specifically have to write that stuff into the component.

- Ran linter

- Added Fix and Non-Fix versions of the lint commands to package.json

- All unit tests passing.


README.md
```diff
@@ -46,24 +46,29 @@ A list of recommended/helpful VS Code extensions.
...
 - Dev: https://public-lng-dev.pathfinder.gov.bc.ca/
 - Test: https://public-lng-test.pathfinder.gov.bc.ca/
 - Prod: TBD
 
-### NRCED
-- Dev: TBD
-- Test: TBD
+### PUBLIC-NRCED
+- Dev: https://public-nrced-dev.pathfinder.gov.bc.ca/
+- Test: https://public-nrpti-test.pathfinder.gov.bc.ca/
```
marklise: This needs to change to public-nrced-test.pathfinder.gov.bc.ca.

PR #87, opened 2020-01-09
PTI-4-2: Update epic import to create initial default flavour records.

NickPhura:
Added a new call to the epic import that creates all of the the flavour records for that master record, if they dont already exist.

### DIFF
--- a/api/src/integrations/epic/epic-orders.js
+++ b/api/src/integrations/epic/epic-orders.js
@@ -3,6 +3,7 @@
 const mongoose = require('mongoose');
 const defaultLog = require('../../utils/logger')('epic-orders');
 const axios = require('axios');
+const INTEGRATION_SCHEMA = require('../integration-schema-enum');
 const hostPath = `https://${process.env.EPIC_API_HOSTNAME || 'eagle-prod.pathfinder.gov.bc.ca'}${process.env
   .EPIC_API_PROJECT_PATHNAME || '/api/project'}/`;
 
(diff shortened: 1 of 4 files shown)


marklise:
What bug number is this for?

api/src/integrations/epic/epic-orders.js
```diff
@@ -95,6 +96,49 @@ class EpicOrders {
...
+
+    try {
+      INTEGRATION_SCHEMA.order.flavourSchemas.forEach(async flavourSchema => {
+        const flavourModel = mongoose.model(flavourSchema);
+
+        const flavourAlreadyExists = await flavourModel.exists({
+          _schemaName: flavourSchema,
+          _master: orderRecord._id
+        });
+
```
danieltruong: A user could add the flavour record before this completes resulting in two records.

NickPhura:
My original thought process behind this, is that the flavours need to exist in order to fetch records properly.  Since we request the flavour, and the master is included via the backref.

But, now that I think about this more:

From the perspective of the public sites - we only need the flavours to exist when the record is Published, so that we can fetch them correctly on the public site.

From the admin side of things, we only need to create the flavours if the admin edits the record and adds flavour values OR when they click publish.

So we probably don't actually need to create the flavours during the import.  The only thing this makes slightly worse, is that when we load the Details or Add/Edit pages, we would potentially have to do 2 queries.  One to fetch the flavours (and master via backref), which may not yet exist, and then in that case a second to directly fetch master.  But, that is probably better than preemptively creating a bunch of flavour records, eh?

api/src/integrations/epic/epic-orders.js
NickPhura: > 
> 
> What bug number is this for?

Not for any bug. See my comment below.

NickPhura:
Determined this is not the right approach after chatting.

PR #88, opened 2020-01-13
[PTI-273]: Data Migration for LNG

marklise:
Data migration of authorizations, compliances, nation agreements, plans, and activities.  Updated static assets from public-lng prod site.  Added
models for a number of NRPTI datasets.


api/src/models/master/plan.js
```diff
@@ -0,0 +1,30 @@
...
+    _schemaName: { type: String, default: 'Plan', index: true },
+    _epicProjectId: { type: 'ObjectId', default: null, index: true },
+    _sourceRefId: { type: 'ObjectId', default: null, index: true },
+    _epicMilestoneId: { type: 'ObjectId', default: null, index: true },
+
+    read: [{ type: String, trim: true, default: 'sysadmin' }],
+    write: [{ type: String, trim: true, default: 'sysadmin' }],
+
+    recordName: { type: String, default: '' },
+    recordType: { type: String, default: '' },
```
NickPhura: Do we still have recordType?  I know there was some confusion before around 'type' since Lena/Kyle assumed it was a field, and we assumed it was the schema.  But maybe this is unrelated to that?

marklise: Unrelated - it's just a field on these particular documents.

PR #89, opened 2020-01-16
PTI-329: update epic import, fix unit tests, add/fix lint commands.

NickPhura:
- All api unit tests passing.
- Updated epic import to move the project fetching into epic-datasource (out of the type specific utils).
  - Partially because I think it fits better there, and partially because it makes the testing easier.
- Added better lint commands to package.json:
  - It now has both non-fix and fix versions, with updated config.
- Added the migrations, and other similar folders, to the lint ignore.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -6,10 +6,12 @@
   "scripts": {
     "start": "node app",
     "test": "node_modules/.bin/jest --verbose",
-    "lint": "eslint . --no-fix --ignore-pattern 'node_modules' --ext .js",
-    "lint-fix:1": "eslint --fix --ignore-pattern 'node_modules' --ext .js",
-    "lint-fix:2": "prettier ./**/*.js --write --loglevel warn",
-    "lint-fix": "npm-run-all -l -s -c lint-fix:*"
+    "lint": "npm-run-all -l -s -c lint:*",
+    "lint:1": "eslint . --ignore-pattern 'node_modules' --ext .js",
+    "lint:2": "prettier --check ./**/*.js ",
+    "lint-fix": "npm-run-all -l -s -c lint-fix:*",
+    "lint-fix:1": "eslint . -fix --ignore-pattern 'node_modules' --ext .js",
+    "lint-fix:2": "prettier ./**/*.js --write --loglevel warn"
   },
   "engines": {
     "node": ">= 10.0.0",
--- a/api/.eslintignore
+++ b/api/.eslintignore
@@ -1,9 +1,10 @@
 node_modules
 
 dist
+coverage
+
 migrations
 migrations_data
-openshift
 
 openshift
 sonar-runner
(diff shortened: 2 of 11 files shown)


api/package.json
```diff
@@ -6,10 +6,12 @@
   "scripts": {
     "start": "node app",
     "test": "node_modules/.bin/jest --verbose",
-    "lint": "eslint . --no-fix --ignore-pattern 'node_modules' --ext .js",
-    "lint-fix:1": "eslint --fix --ignore-pattern 'node_modules' --ext .js",
-    "lint-fix:2": "prettier ./**/*.js --write --loglevel warn",
-    "lint-fix": "npm-run-all -l -s -c lint-fix:*"
+    "lint": "npm-run-all -l -s -c lint:*",
+    "lint:1": "eslint ./**/*.js",
```
marklise: npm run lint:1 is failing

NickPhura: The command is failing, or the linter is reporting lint errors?

marklise: ```npm run lint``` in the api directory, you'll see the error.

NickPhura: Locally, the command is working for me, and it reports a bunch of linting errors.  I didnt fix them as part of not linting code that doesnt have changes in it though.

NickPhura: Im not quite sure what error you're seeing with the linting.  For me locally, the command runs, reports linting errors, and ultimately reports that it failed, but only because of the linting issues.  The command itself executed fully, as intended.

Are you seeing an actual error that prevents the lint commands from running at all?

marklise: image

marklise: image

NickPhura: Hmm, not sure what thats about.

The pattern error makes no sense to me.  Why would it have anything to do with node_modules, let alone as the pattern.  node_modules is also in the ignore file, as usual.

NickPhura: I wonder if it has to do with a difference in shell or OS?

NickPhura: Is there a version of that command that works for you locally? Maybe you need brackets around the pattern?

If you change it to:

```
"lint:1": "eslint \"./**/*.js\"",
```

does it work for you?

NickPhura: Any luck?

marklise: In the screenshot it shows that you're running the linter against the node_modules directory.  Even though the eslintignore is there, it will catch any directory that has a .js in it's name.

NickPhura: I can't seem to reproduce this issue, so I'm not sure how to resolve this.  

The .eslintignore seems to be working correctly for me, as when I add '--no-ignore' it tries to lint the migration files that it would normally skip.  Even with the ignore commented out entirely, it still doesn't try to lint node_modules, because I think eslint ignores it by default.

PR #90, opened 2020-01-16
PTI-311: Universal Post

danieltruong:


api/src/controllers/post/inspection.js
```diff
@@ -0,0 +1,178 @@
...
+
+    // We must have a valid master ObjectID to continue.
+    masterId && ObjectId.isValid(masterId) && (inpsectionLNG._master = new ObjectId(masterId));
+    if (!inpsectionLNG._master) {
+        return {
+            status: 'failure',
+            object: incomingObj,
+            errorMessage: 'incomingObj._master was not valid ObjectId'
+        }
+    }
```
NickPhura: Minor optimization, but the check for a valid masterId could be the first thing this function checks, before loading the model, etc.

api/src/controllers/post/inspection.js
```diff
@@ -0,0 +1,178 @@
...
+ *      description: 'nrced description',
+ *      ...
+ *      addRole: 'public'
+ *  }
+ */
+exports.createNRCED = async function (args, res, next, incomingObj, masterId) {
+    var InspectionNRCED = mongoose.model('InspectionNRCED');
+    var inpsectionNRCED = new InspectionNRCED();
+
+    inpsectionNRCED._schemaName = 'InspectionNRCED';
```
NickPhura: Not required, but its not a bad idea to save strings like 'InspectionNRCED', and similar, as constants at the top level.  That way its easy to update every instance of its use at once, without having to track them all down.  Could use it in the mongoose.model() call as well, etc.

If these schema names are used across multiple files, then they could also be moved into a separate enum or constants file, so theres only 1 place that needs to be updated across the entire api.

api/src/controllers/post/inspection.js
```diff
@@ -0,0 +1,178 @@
...
+    incomingObj.sourceDateUpdated && (inpsection.sourceDateUpdated = incomingObj.sourceDateUpdated);
+    incomingObj.sourceSystemRef && (inpsection.sourceSystemRef = incomingObj.sourceSystemRef);
+
+    inpsection.read = ['sysadmin'];
+    inpsection.write = ['sysadmin'];
+
+    let savedInspection = null;
+    try {
+        savedInspection = await inpsection.save();
+    } catch (e) {
```
NickPhura: I'd prefer 'error' or at least 'err' over just 'e'

Helps with readability, and also makes searching better.  Its pretty impractical to search for 'e', if you were trying to find places where errors are handled and weren't familiar with the code base.

api/src/controllers/post/inspection.js
```diff
@@ -0,0 +1,178 @@
+var mongoose = require('mongoose');
+var ObjectId = require('mongoose').Types.ObjectId;
+
+// Example of incomingObj
```
NickPhura: nitpick:  You might as well include this comment inside the /** */ comment block, rather than have it separated.

api/src/controllers/record-controller.js
```diff
@@ -433,4 +489,34 @@ let addStandardQueryFilters = function(query, args) {
 
   return query;
 };
+
+let processPostRequest = async function (args, res, next, property, data) {
+  var i = data.length - 1;
+  var observables = [];
+
+  do {
```
NickPhura: With the do-while loop, we probably need to check that 'i' is > 0.  Otherwise your first loop would try and access data[-1]

NickPhura:
Added some comments.  

I didn't post the comments twice (on both Order and Inspection code) but they apply to of course since its the same logic.

api/src/controllers/post/inspection.js
danieltruong: You could always search for "catch ("

PR #98, opened 2020-01-17
PTI-258: Order/Inspection admin detail page. 

NickPhura:
## Changes
Add Order and Inspection admin details pages.
 - Order and Inspection details pages are fairly identical.
   - Currently the flavour details are separate components that are injected into the master details component (if data for them is returned from the api).
   - Future optimization: Because they are so similar, it would make sense to move the scss, and maybe some of the template, into shared files.  But for now, for the sake of the demo, they just all have the same code separately, as breaking scss into fancier classes will take me some time.  I also need to sync up with which classes (for font, and color, etc) are standard with the new Eagle scss.  With Daniels  add/edit pages coming along, I can just take inspiration from those later.

Wire up publish/unpublish.
 - There are generic `record/{objectId}/publish` and `record/{objectId}/unpublish` swagger routes.
 - Currently this calls a standalone handler in the record-controller, which uses the publish/unpublish actions.
   - Eventually this could just call into Daniels universal put/post stuff.


angular/projects/admin-nrpti/src/app/records/inspections/inspection-detail/inspection-detail.component.html
```diff
@@ -0,0 +1,100 @@
+<main class="detail-page">
+  <div>
+    <section class="content-section">
+      <div class="content-grid">
+        <div class="grid-section-header">
+          <div class="grid-section-header-title">
+            <span class="grid-section-header-title-text">Shared Data [Master]</span>
+            <span class="grid-section-header-title-sub-text"
+              >Last Edited {{ (data && data._master.dateUpdated | date: 'mediumDate') || '-' }}</span
```
marklise: Can we close > on within the same tag. Same for all occurrences below.

NickPhura: Done.

I also updated the formatter to default to that style.

NickPhura:
> 
> 
> Order Edit screen bombs. Click on edit button in the details page of an Order.

Yeah it looks like the edit page is just slightly outdated, so its not expecting the right response from the resolver. 

I just commented out the edit button routing for now, while the edit page is a WIP.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
```diff
@@ -0,0 +1,96 @@
...
+@Component({
+  selector: 'app-order-lng-detail',
+  templateUrl: './order-lng-detail.component.html',
+  styleUrls: ['./order-lng-detail.component.scss']
+})
+export class OrderLNGDetailComponent extends RecordComponent implements OnInit, OnDestroy {
+  private ngUnsubscribe: Subject<boolean> = new Subject<boolean>();
+
+  public isLoading = true;
+  public isPublishing = false;
```
marklise: This is one of many places where isPublishing is not used in the UI at all. Should be tossed.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  publish(): void {
+    this.isPublishing = true;
+
+    this.factoryService
+      .publishRecord(this.data)
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(response => {
+        this.data = new OrderLNG(response);
+
```
marklise: If you're not reloading the page, you should use changeDetection incase there are other fields that have updated since originally fetching.  We should account for error cases (already published).

Same goes for all other areas in the diff.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
```diff
@@ -0,0 +1,96 @@
...
+      .unPublishRecord(this.data)
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe(response => {
+        this.data = new OrderLNG(response);
+
+        this.isUnPublishing = false;
+      });
+  }
+
+  isPublished(): boolean {
```
marklise: Instead of calling isPublished() on an *ngIf, this should be calculated/cached at ngOnInit() time or re-calculated and set to a variable right before doing change detection.  Same goes for other functions inside the *ngIf.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-detail/inspection-detail.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@Component({
+  selector: 'app-inspection-detail',
+  templateUrl: './inspection-detail.component.html',
+  styleUrls: ['./inspection-detail.component.scss']
+})
+export class InspectionDetailComponent extends RecordComponent implements OnInit, OnDestroy {
+  private ngUnsubscribe: Subject<boolean> = new Subject<boolean>();
+
+  public isLoading = true;
```
marklise: This is one of many places where isLoading is not used in the UI at all. Should be tossed.  The only place that uses this paradigm is the map.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-detail/inspection-detail.component.html
```diff
@@ -0,0 +1,100 @@
...
+    <section class="content-section">
+      <div class="content-grid">
+        <div class="grid-section-header">
+          <div class="grid-section-header-title">
+            <span class="grid-section-header-title-text">Shared Data [Master]</span>
+            <span class="grid-section-header-title-sub-text">
+              Last Edited {{ (data && data._master.dateUpdated | date: 'mediumDate') || '-' }}
+            </span>
+          </div>
+          <div class="grid-section-header-button">
```
marklise: Remove this entirely if there is no edit functionality.  Same goes for the Order.

api/src/controllers/record-controller.js
```diff
@@ -212,7 +197,7 @@ exports.protectedUnPublish = function (args, res, next) {
  * @param {*} next
  * @returns
  */
-exports.publicHead = function (args, res, next) {
+exports.publicHead = function(args, res, next) {
```
marklise: We should get rid of these since we don't support publicHead.  Also - keep whitespace changes for other commits unless you're fixing logic in said function to keep reviews to a minimum.

api/src/controllers/record-controller.js
```diff
@@ -176,30 +117,74 @@ exports.protectedPut = function (args, res, next) {
...
  */
-exports.protectedUnPublish = function (args, res, next) {
-  return queryActions.sendResponse(res, 501);
+exports.protectedUnPublish = function(args, res, next) {
+  const recordData = args.swagger.params.record.value;
+  defaultLog.info(`protectedUnPublish - recordId: ${recordData._id}`);
+
+  const model = require('mongoose').model(recordData._schemaName);
+
+  model.findOne({ _id: recordData._id }, async function(error, record) {
```
marklise: Convert to async/await.

api/src/controllers/record-controller.js
```diff
@@ -176,30 +117,74 @@ exports.protectedPut = function (args, res, next) {
...
  */
-exports.protectedPublish = function (args, res, next) {
-  return queryActions.sendResponse(res, 501);
+exports.protectedPublish = function(args, res, next) {
+  const recordData = args.swagger.params.record.value;
+  defaultLog.info(`protectedPublish - recordId: ${recordData._id}`);
+
+  const model = require('mongoose').model(recordData._schemaName);
+
+  model.findOne({ _id: recordData._id }, async function(error, record) {
```
marklise: Convert to async/await.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-lng-detail/inspection-lng-detail.component.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      this.isLoading = false;
+    });
+  }
+
+  publish(): void {
+    this.isPublishing = true;
+
+    this.factoryService
+      .publishRecord(this.data)
```
marklise: These throw, and need to be caught.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
NickPhura: Not 100% sure if they way I changed it is what you had in mind, let me know.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-detail/inspection-detail.component.ts
NickPhura: Removed

api/src/controllers/record-controller.js
NickPhura: Removed.

api/src/controllers/record-controller.js
NickPhura: Fixed.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-lng-detail/inspection-lng-detail.component.ts
NickPhura: Added catch pipe step + handling.

api/src/controllers/record-controller.js
NickPhura: Fixed.

angular/projects/admin-nrpti/src/app/records/inspections/inspection-detail/inspection-detail.component.html
NickPhura: Commented out for now while its not functional.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
NickPhura: Added better error handling for thrown errors, and already published/unpublished cases.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
NickPhura: Removed.

angular/projects/admin-nrpti/src/app/records/orders/order-lng-detail/order-lng-detail.component.ts
marklise: This is better using a variable, however one last little bit is missing.  Here and in other places you implemented ngOnChanges() which is not quite what we've normally been doing.  That function is called when changes are changed in input based data bound variables.  Once you change data bound props inside the subscribe eg: `this.foo = 'bar'`, you need to tell angular to run the change detection cycle as it's possible that the service broke out of angular's zone.  Remove the ngOnChanges, and instead Import changeDetection and call the detectChanges() function on it, like so: `this._changeDetectionRef.detectChanges();`

See `inspection-detail.component.ts` in public-nrpti project for an example.

NickPhura: Ah right, I forgot about that import.

Cool, fixed.

I also moved the status string to a variable, instead of a function referenced in the template.

PR #99, opened 2020-01-17
 PTI-238: Add orders

danieltruong:


angular/projects/admin-nrpti/src/app/records/orders/order-add-edit/order-add-edit.component.html
```diff
@@ -27,8 +27,10 @@ <h4 class="blue-header">Shared Data [Master]</h4>
           <div class="row">
             <div class="col-7"></div>
             <div class="col-5">
-              <span class="grey-subtext">If content is published on any site, hitting the Save/Update
-                Record button will update the public site immediately</span>
+              <span class="grey-subtext"
+                >If content is published on any site, hitting the Save/Update Record button will update the public site
```
marklise: Put the > on the line above

NickPhura: I dont know all of the ins and outs of spans, but I believe the reason for the > on the lower line is to ensure that intentional spaces that might come before the text aren't confused with the indentation spaces from the code formatting.

angular/tslint.json
```diff
@@ -89,7 +89,7 @@
     "no-submodule-imports": false,
     "no-switch-case-fall-through": true,
     "no-trailing-whitespace": true,
-    "no-unused-expression": true,
+    "no-unused-expression": false,
```
NickPhura: Why unset this setting?

If there are variables, etc, that are completely unused, its best to remove them so they don't  clutter the code unnecessarily.

angular/projects/global/src/lib/utils/utils.ts
```diff
@@ -22,4 +30,90 @@ export class Utils {
...
+        .replace(/\%2F/g, '_')
+        .replace(/ /g, '_');
+    } else {
+      return filename
+        .replace(/\(/g, '%28')
+        .replace(/\)/g, '%29')
+        .replace(/\\/g, '_')
+        .replace(/\//g, '_');
+    }
+  }
```
NickPhura: Is there any reason why we wouldn't just use something like: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

to handle encoding?

angular/projects/global/src/lib/utils/utils.ts
```diff
@@ -22,4 +30,90 @@ export class Utils {
...
+        month = '0' + month;
+      }
+      if (day.length < 2) {
+        day = '0' + day;
+      }
+      return [year, month, day].join('');
+    } else {
+      return null;
+    }
+  }
```
NickPhura: This seems like the kind of stuff MomentJS is good at.  Or is there a reason to do it manually ourselves?

angular/projects/global/src/lib/utils/utils.ts
```diff
@@ -1,3 +1,11 @@
+const encode = encodeURIComponent;
+window['encodeURIComponent'] = (component: string) => {
+  return encode(component).replace(/[!'()*]/g, c => {
+    // Also encode !, ', (, ), and *
+    return '%' + c.charCodeAt(0).toString(16);
+  });
+};
+
```
NickPhura: Similar question further below.

Any reason not to use something like: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

to handle encoding?

angular/tslint.json
danieltruong: This was causing short circuiting to fail lint

angular/projects/global/src/lib/utils/utils.ts
marklise: This should be tossed as it's not being used anywhere.

angular/projects/global/src/lib/utils/utils.ts
marklise: Toss this as it's not being used.

angular/tslint.json
NickPhura: Apparently they added a setting to account for this: https://palantir.github.io/tslint/rules/no-unused-expression/

So we just need to change it to:

```"no-unused-expression": [true, "allow-fast-null-checks"]```

PR #106, opened 2020-01-21
PTI-311: Added universal put

danieltruong:


api/src/controllers/put/inspection.js
```diff
@@ -0,0 +1,232 @@
...
+ */
+exports.editMaster = async function (args, res, next, incomingObj) {
+    var _id = null;
+    if (!incomingObj._id) {
+        return {
+            status: 'failure',
+            object: incomingObj,
+            errorMessage: 'No _id provided'
+        }
+    } else {
```
NickPhura: nitpick: dont need the else wrapper, because you return in the if anyways. 

Same for the cases further down.

PR #111, opened 2020-01-21
PTI-238: Edit order interface for NRPTI admin

danieltruong:


angular/projects/admin-nrpti/src/app/records/orders/order-add-edit/order-add-edit.component.ts
```diff
@@ -45,8 +49,9 @@ export class OrderAddEditComponent implements OnInit, OnDestroy {
     this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
       !(res.breadcrumb === 'Add Order') && (this.isEditing = true);
```
NickPhura: Since isEditing is always false or true, this can be simplified a bit:
```
this.isEditing = !(res.breadcrumb === 'Add Order');
```

angular/projects/admin-nrpti/src/app/records/orders/order-add-edit/order-add-edit.component.ts
```diff
@@ -147,33 +184,49 @@ export class OrderAddEditComponent implements OnInit, OnDestroy {
       summary: this.myForm.controls.nrcedSummary.value
     };
     this.nrcedPublishStatus === 'Publish' && (order.OrderNRCED['addRole'] = 'public');
+    this.isEditing && this.nrcedPublishStatus === 'Unpublish' && (order.OrderNRCED['removeRole'] = 'public');
 
     order.OrderLNG = {
       description: this.myForm.controls.lngDescription.value
     };
     this.lngPublishStatus === 'Publish' && (order.OrderLNG['addRole'] = 'public');
+    this.isEditing && this.lngPublishStatus === 'Unpublish' && (order.OrderLNG['removeRole'] = 'public');
```
NickPhura: Similar to my other comment, and maybe this is subjective, but I think its a bit awkward to set a value in the middle of a long chain like this, especially because at first glance you don't really notice it is setting a value, until you realize its a single equal and not a double or triple.

I think it reads a lot cleaner the traditional way:
```
if(this.isEditing && (this.lngPublishStatus === 'Unpublish')) {
  order.OrderLNG['removeRole'] = 'public';
}
```

api/src/controllers/put/inspection.js
```diff
@@ -74,6 +70,20 @@ exports.editMaster = async function (args, res, next, incomingObj) {
     }
   }
 
+  // Flavours:
+  // When editing, we might get a request to make a brand new flavour rather than edit.
+  var observables = [];
+  if (incomingObj.InspectionLNG && incomingObj.InspectionLNG._id) {
+    observables.push(this.editLNG(args, res, next, incomingObj.InspectionLNG)) && delete incomingObj.InspectionLNG;
```
NickPhura: Is there any reason why these 2 commands are being chained?

Because observables.push() is never going to fail, so its always going to run the delete anyways.  Unless I am missing something, they might as well just be separate lines, no?
```
if (...) {
  observables.push(this.editLNG(args, res, next, incomingObj.InspectionLNG));
  delete incomingObj.InspectionLNG;
}
```

Same comment for the other instances in the PR.

PR #112, opened 2020-01-21
PTI-258-3: Improve Publish/Unpublish logic + UI date handling.

NickPhura:
- Updated pub/unpub + UI handling

- Also included an update to the record-detail.directive because I noticed it was missing.  Matches the behaviour of the table-row.directive now to handle the input more correctly.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/inspections/inspection-lng-detail/inspection-lng-detail.component.ts
+++ b/angular/projects/admin-nrpti/src/app/records/inspections/inspection-lng-detail/inspection-lng-detail.component.ts
@@ -16,7 +16,6 @@ export class InspectionLNGDetailComponent extends RecordComponent implements OnI
   private ngUnsubscribe: Subject<boolean> = new Subject<boolean>();
 
   public isPublished: boolean;
-  public publishStatusString: string;
 
   constructor(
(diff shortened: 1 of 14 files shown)


angular/projects/admin-nrpti/src/app/records/inspections/inspection-lng-detail/inspection-lng-detail.component.ts
```diff
@@ -114,7 +114,13 @@ export class InspectionLNGDetailComponent extends RecordComponent implements OnI
       return 'Not Published';
     }
 
-    return `Published on ${this.datePipe.transform(this.data.datePublished, 'mediumDate')}`;
+    const publishDate = this.datePipe.transform(this.data && this.data.datePublished, 'mediumDate');
```
marklise: I'm not sure why we're doing logic here, instead of pushing it into a pipe in the template.

NickPhura: Just because the logic in this function is probably a bit complex to put all inline in the template.

It does checks on this.data && this.data.datePublished to ensure its not null, and then returns one of 3 strings.

PR #123, opened 2020-01-23
PTI-345: Hide edit field if dateUpdated is not populated

danieltruong:


danieltruong:
That was a separate issue. The migration script simply didn't set the fields when migrating. This issue was somewhat a non-issue. When editing a record that does not have a flavour, filling in a flavour's details will create the flavour. Therefore, it uses post rather than put. This means dateUpdated does not get set which is expected. This change accounts for that and just hides the text until someone actually updates an existing flavour.

PR #134, opened 2020-01-28
PTI:114: Adding filters to authorizations, back-end changes, fix overflow chop…

marklise:
…ping off calendar UI

### DIFF
--- a/angular/projects/public-lng/src/app/explore-panel/date-input/date-input.component.ts
+++ b/angular/projects/public-lng/src/app/explore-panel/date-input/date-input.component.ts
@@ -1,0 +1,49 @@
+import { Component, Input, Output, EventEmitter, OnChanges, SimpleChanges } from '@angular/core';
+import { NgbDateStruct } from '@ng-bootstrap/ng-bootstrap';
+import { Utils } from 'nrpti-angular-components';
+
+@Component({
+  selector: 'app-date-input',
+  templateUrl: './date-input.component.html',
+  styleUrls: ['./date-input.component.scss']
+})
+export class DateInputComponent implements OnChanges {
+  @Input() date: Date     = null;
+  @Input() isValidate     = false;
+  @Input() minDate: Date  = null;
+  @Input() maxDate: Date  = null;
+  @Output() dateChange    = new EventEmitter<any>();
+
+  public ngbDate: NgbDateStruct     = null;
+  public minNgbDate: NgbDateStruct  = null;
+  public maxNgbDate: NgbDateStruct  = null;
+
+  constructor(
+    private utils: Utils,
+  ) {}
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.date && changes.date.currentValue) {
+      this.ngbDate = this.utils.convertJSDateToNGBDate(new Date(changes.date.currentValue));
+    }
+    if (changes.minDate && changes.minDate.currentValue) {
+      this.minNgbDate =  this.utils.convertJSDateToNGBDate(new Date(changes.minDate.currentValue));
+    }
+    if (changes.maxDate && changes.maxDate.currentValue) {
+      this.maxNgbDate =  this.utils.convertJSDateToNGBDate(new Date(changes.maxDate.currentValue));
+    }
+  }
+
+  onDateChg(ngbDate: NgbDateStruct) {
+    this.dateChange.emit(ngbDate ? this.utils.convertFormGroupNGBDateToJSDate(this.ngbDate).toISOString() : null);
+  }
+
+  clearDate() {
+    this.ngbDate = null;
+    this.onDateChg(this.ngbDate);
+  }
+
+  public isValidDate(date: NgbDateStruct): boolean {
+    return date && !isNaN(date.year) && !isNaN(date.month) && !isNaN(date.day);
+  }
+}
--- a/angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
+++ b/angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
@@ -1,0 +1,123 @@
+@import "~assets/styles/base/base.scss";
+
+.form-group {
+  margin-bottom: 2rem;
+}
+
+.form-btns {
+  .btn-primary {
+    font-weight: 700;
+  }
+
+  .btn-light {
+    font-weight: 400;
+  }
+}
+
+// Date Range Inputs
+.date-range {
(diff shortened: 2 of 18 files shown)


angular/projects/public-lng/src/app/explore-panel/date-input/date-input.component.ts
```diff
@@ -0,0 +1,49 @@
+import { Component, Input, Output, EventEmitter, OnChanges, SimpleChanges } from '@angular/core';
+import { NgbDateStruct } from '@ng-bootstrap/ng-bootstrap';
+import { Utils } from 'nrpti-angular-components';
+
+@Component({
+  selector: 'app-date-input',
+  templateUrl: './date-input.component.html',
+  styleUrls: ['./date-input.component.scss']
+})
+export class DateInputComponent implements OnChanges {
```
NickPhura: This component seems like a prime candidate for the global library.

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
```diff
@@ -0,0 +1,123 @@
...
+
+// show/hide button variants for large and small devices
+@media (max-width: 1023px) {
+  .explore-btn-lg-device {
+    display: none;
+  }
+}
+
+@media (min-width: 1024px) {
+  .explore-btn-sm-device {
```
NickPhura: This class isn't used anywhere, so when the screen is resized smaller the 'Apply' date-picker button disappears entirely.

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  constructor(
+    private _changeDetectionRef: ChangeDetectorRef,
+    private route: ActivatedRoute
+    ) {
+      this.textFilterKeys = [];
+    }
+
+  public ngOnInit() {
+    // For each filter coming in, make the filter bucket array
+    this.route.params.subscribe(params => {
```
NickPhura: I'm not actually clear on how important ```takeUntil(this.ngUnsubscribe)``` is, but something that could be added if needed.

angular/projects/public-lng/src/app/project/authorizations/authorizations.component.ts
```diff
@@ -25,12 +15,70 @@ export class AuthorizationsComponent implements OnInit {
...
+        if (Object.keys(event).includes(p)) {
+          // We will be overriding this param later.
+        } else {
+          // Existing param we should save.
+          newParams[p] = params[p];
+        }
+      });
+
+      Object.keys(event).forEach(item => {
+        if (event[item] === undefined || event[item].length === 0) {
```
NickPhura: ```event[item].length``` will error if ```event[item]``` is null.

angular/projects/public-lng/src/app/project/authorizations/authorizations.component.ts
```diff
@@ -25,12 +15,70 @@ export class AuthorizationsComponent implements OnInit {
...
+          // console.log('skipping:', e);
+        } else {
+          newParams[item] = event[item];
+        }
```
NickPhura: Instead of having an empty if(), you could do:

```
If (event[item] !== undefined && event[item] !== null && event[item].length > 0 {
newParams[item] = event[item]
```

api/src/controllers/search.js
```diff
@@ -291,6 +258,89 @@ exports.protectedGet = function (args, res, next) {
...
+  }
+
+  var masterModifier = {};
+  if (andMasterExpArray.length > 0 && orMasterExpArray.length > 0) {
+    masterModifier = { $and: [{ $and: andMasterExpArray }, { $and: orMasterExpArray }] };
+  } else if (andMasterExpArray.length === 0 && orMasterExpArray.length > 0) {
+    masterModifier = { $and: orMasterExpArray };
+  } else if (andMasterExpArray.length > 0 && orMasterExpArray.length === 0) {
+    masterModifier = { $and: andMasterExpArray };
+  }
```
NickPhura: Shouldn't at least some of these use ```$or``` ?

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
NickPhura: This was supposed to be the comment that needed changes, not the $or question further below.

api/src/controllers/search.js
NickPhura: Shouldn't at least some of these use $or ?

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
NickPhura: This class isn't used anywhere, so when the screen is resized smaller the 'Apply' date-picker button disappears entirely.

angular/projects/public-lng/src/app/explore-panel/date-input/date-input.component.ts
marklise: As mentioned in the stand up yesterday, I was to land this as is.

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.ts
marklise: It's an error, will fix.

angular/projects/public-lng/src/app/project/authorizations/authorizations.component.ts
marklise: event will never be null as it's called from a change event in the explorer component, however not a bad extra sanity check.

angular/projects/public-lng/src/app/project/authorizations/authorizations.component.ts
marklise: It's part comment and part left for debugging.  I don't think it hurts anything.

api/src/controllers/search.js
marklise: The logic is that you $and the split elements coming from the API call, and if they are arrays, you'll see in the generate EXP they get constructed as an $or.  It should all work.

angular/projects/public-lng/src/app/explore-panel/explore-panel.component.scss
marklise: Imported css - will fix.

PR #140, opened 2020-01-30
PTI-292: Add Agreement details and add/edit.

NickPhura:
Add agreement models.
Add agreement add/edit pages.
Add agreement details page.

Minor update to lng migrations (change agreement ```date``` to ```dateIssued``` to align with other record models).

### DIFF
--- a/api/migrations/20200109220142-projectData.js
+++ b/api/migrations/20200109220142-projectData.js
@@ -160,7 +160,7 @@ let createAgreementRecord = async function (item, project, nrptiCollection) {
     recordName: item.name,
     recordType: item.complianceDocumentType,
     // Prefer to store dates in the DB as ISO, not some random format.
-    date: moment(item.date, 'DD-MM-YYYY').toDate(),
+    dateIssued: moment(item.date, 'DD-MM-YYYY').toDate(),
     nationName: item.nation,
     attachments: [{ url: item.url }],
     projectName: project === '588511c4aaecd9001b825604' ? 'LNG Canada' : 'Coastal Gaslink',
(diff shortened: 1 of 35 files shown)


api/migrations/20200109220142-projectData.js
```diff
@@ -160,7 +160,7 @@ let createAgreementRecord = async function (item, project, nrptiCollection) {
     recordName: item.name,
     recordType: item.complianceDocumentType,
     // Prefer to store dates in the DB as ISO, not some random format.
-    date: moment(item.date, 'DD-MM-YYYY').toDate(),
+    dateIssued: moment(item.date, 'DD-MM-YYYY').toDate(),
```
marklise: This breaks public LNG site.  You'll need to change the date handling for agreement documents over there as well.

NickPhura: Fixed in second commit. Ran the migration and checked that it matches what shows up in dev.

api/src/models/master/agreement.js
```diff
@@ -12,8 +12,10 @@ module.exports = require('../../utils/model-schema-generator')(
     write: [{ type: String, trim: true, default: 'sysadmin' }],
 
     recordName: { type: String, default: '' },
-    date: { type: Date, default: Date.now() },
+    recordType: { type: String, default: '' },
+    dateIssued: { type: Date, default: Date.now() },
```
marklise: We shouldn't bother with Date.now(), better to leave it blank otherwise we are 'creating' things that don't exist.

NickPhura: Fixed.  Ill have to go back and update the other models as well then after.

PR #174, opened 2020-02-13
Reduce epic-import parallelization.

NickPhura:
### Previous behaviour:
- For each supported type, run updates in parallel.
  - For each record of that type, run update in parallel.

### Updated behaviour:
- For each supported type, run updates in parallel.
  - For each record of that type, run update in batches of 50.

#### I didn't update the project URL in this PR, because it sounds like there is some controversy around that issue still.


NickPhura:
Replacing with PR from branch (rather than master)

PR #185, opened 2020-02-19
PTI-257: Updated document architecture

danieltruong:
- Moved shared module to its own directory in admin.
- Moved file upload to common.
- Added scss to common for file upload build and test to work.
- Records now have a field called documents which is an array of object ids that correspond to documents.
- We now have document objects.
- Updated import from EPIC to account for how we do documents now.

Note:
- I have included initial work for uploading into S3. These functions have been marked with a WIP.


api/src/controllers/document-controller.js
```diff
@@ -0,0 +1,113 @@
...
+const mongoose = require('mongoose');
+const queryActions = require('../utils/query-actions');
+const { uuid } = require('uuidv4');
+const AWS = require('aws-sdk');
+
+const ep = new AWS.Endpoint('nrs.objectstore.gov.bc.ca');
+const s3 = new AWS.S3(
+  {
+    endpoint: ep,
+    accessKeyId: 'THIS IS WHERE THE ACCESS KEY GOES',
```
marklise: These need to pull from an environment variable.

api/src/controllers/document-controller.js
```diff
@@ -0,0 +1,113 @@
...
+  res.status(200).send();
+};
+
+// WIP
+exports.protectedPut = async function (args, res, next) {
+  if (args.swagger.params.data && args.swagger.params.data.value) {
+    const data = args.swagger.params.data.value;
+
+    if (!data.fileName) {
+      return queryActions.sendResponse(res, 500, 'You must have a file name');
```
marklise: 400, not 500

api/src/controllers/document-controller.js
```diff
@@ -0,0 +1,113 @@
...
+  document.url = url;
+  document.read = ['public', 'sysadmin'];
+  document.write = ['sysadmin'];
+  return await document.save();;
+}
+
+// WIP
+function createSignedUrl(operation, key) {
+  let params = {
+    Bucket: 'THIS IS WHERE THE BUCKET GOES',
```
marklise: Environment variable pulled in at the top fo the file and referenced here.

angular/projects/admin-nrpti/src/app/shared/shared.module.ts
```diff
@@ -3,17 +3,17 @@ import { BrowserModule } from '@angular/platform-browser';
...
   imports: [BrowserModule, MatSlideToggleModule, MatSnackBarModule],
-  declarations: [OrderByPipe, NewlinesPipe, ObjectFilterPipe, LinkifyPipe, FileUploadComponent],
+  declarations: [OrderByPipe, NewlinesPipe, ObjectFilterPipe, LinkifyPipe],
   providers: [DatePipe],
-  exports: [MatSlideToggleModule, MatSnackBarModule, OrderByPipe, NewlinesPipe, LinkifyPipe, FileUploadComponent]
+  exports: [MatSlideToggleModule, MatSnackBarModule, OrderByPipe, NewlinesPipe, LinkifyPipe],
+  entryComponents: []
 })
-export class SharedModule {}
+export class SharedModule { }
```
NickPhura: Not a priority issue: maybe just double check your vscode formatter settings, its still adding a space here when it shouldn't.  I thought we found and fixed that setting a while back, but maybe that was only for the function spacing.

angular/projects/common/src/app/models/document.ts
```diff
@@ -6,20 +6,17 @@
...
-  _record: string;
   _addedBy: string;
-  documentFileName: string;
-  displayName: string;
-  internalURL: string;
-  internalMime: string;
+  fileName: string;
+  key: string;
+  url: string;
+  upfile: File;
```
NickPhura: This value isn't used anywhere here.  Should it be? Or can it be removed? (Unless it gets used in a later update).

api/src/controllers/document-controller.js
```diff
@@ -0,0 +1,113 @@
+
+const mongoose = require('mongoose');
+const queryActions = require('../utils/query-actions');
+const { uuid } = require('uuidv4');
+const AWS = require('aws-sdk');
+
+const ep = new AWS.Endpoint('nrs.objectstore.gov.bc.ca');
```
NickPhura: Probably want to store this as an env var as well.

angular/projects/common/src/app/models/document.ts
danieltruong: It is used. but only for the front end. Its used to hold the binary. we dont actually want to send this over to the api

NickPhura: Maybe just add a comment then.
The linter might complain as well, so we may need to add a lint-ignore comment.

api/src/integrations/epic/inspections-utils.js
```diff
@@ -41,15 +42,17 @@ class Inspections {
...
+        (this.auth_payload && this.auth_payload.displayName) || '',
+        `https://projects.eao.gov.bc.ca/api/document/${epicRecord._id}/fetch/${encodeURIComponent(
           epicRecord.documentFileName
-        )}`,
-        fileName: epicRecord.documentFileName
-      });
+        )}`
+      );
+      documents.push(savedDocument);
     }
```
NickPhura: It would be good to pull this out into epic-utils.js, since it will get used in all of the epic imports.

marklise:
For the deploymentconfig, instead of ENV vars to spec, we need to add the following:

`"envFrom": [
                  {
                    "prefix": "OBJECTSTORE_",
                    "secretRef": {
                      "name": "name-of-secret-that-holds-accesskey_secretkey"
                    }
                  }
                ],`

api/src/integrations/epic/inspections-utils.js
danieltruong: it is pulled out. It's using a function in the DocumentController

PR #194, opened 2020-02-24
PTI-7: NRIS migration

marklise:
Stubs in the first cut of NRIS import task.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/import/import.component.ts
+++ b/angular/projects/admin-nrpti/src/app/import/import.component.ts
@@ -1,5 +1,4 @@
 import { Component, OnInit, ChangeDetectorRef } from '@angular/core';
-import { Observable } from 'rxjs/Observable';
 import { FactoryService } from '../services/factory.service';
 import { Subject } from 'rxjs';
 import {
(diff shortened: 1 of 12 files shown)


angular/projects/admin-nrpti/src/app/import/import.component.ts
```diff
@@ -143,18 +142,19 @@ export class ImportComponent implements OnInit {
 
   checkChange() {}
 
-  startJob() {
+  async startJob() {
+    await this.startTask();
+    this.showAlert = true;
     // tslint:disable-next-line: no-this-assignment
     const self = this;
```
NickPhura: Is the ```self=this``` necessary here?

Since arrow functions don't have their own scope wouldn't ```this.showAlert``` use the 'this' of the current import class?

api/src/integrations/epic/datasource.js
```diff
@@ -25,16 +25,22 @@ class DataSource {
     this.status = { itemsProcessed: 0, itemTotal: 0, typeStatus: [], individualRecordStatus: [] };
   }
 
+  // This requires to auth setup, so just call the local updater function.
```
NickPhura: nit: I assume you meant 'requires **no** auth setup'

api/src/integrations/nris/datasource.js
```diff
@@ -0,0 +1,236 @@
...
+   * @param {*} [params=null] params to filter records (optional).
+   * @param {*} [recordTypes=null] specific record types to update (optional).
+   * @memberof NrisDataSource
+   */
+  constructor(auth_payload, params = null, recordTypes = null) {
+    this.auth_payload = auth_payload;
+    this.params = params || {};
+
+    // Set initial status
+    this.status = { itemsProcessed: 0, itemTotal: 0, typeStatus: [] };
```
NickPhura: ```this.status``` isn't used anywhere

api/src/integrations/nris/datasource.js
```diff
@@ -0,0 +1,236 @@
...
+        message: 'Job Complete',
+        itemsProcessed: 0,
+        itemTotal: 0,
+      };
+
+      // Keep going until we'd start past today's date.
+      while(startDate < moment()) {
+        defaultLog.info("dateRange:", startDate.format('YYYY-MM-DD'), endDate.format('YYYY-MM-DD'));
+        startDate = endDate;
+        endDate = moment(startDate).add(1, 'M');
```
NickPhura: These dates should increment at the end of the while loop.
Currently, the first month of 2017 is being missed.

api/src/integrations/epic/datasource.js
```diff
@@ -58,7 +64,10 @@ class DataSource {
       }
 
       await Promise.all(promises);
+      this.status = { ...this.status, ...this.getStatusTotals() };
```
NickPhura: I think something got messed up during rebase for this file.

The current version in master is different.  It does batch requests to EPIC (to avoid blowing them up).
And for example, getStatusTotals() no longer exists in master, and doesn't exist here either (not sure why its not throwing an error).

PR #197, opened 2020-02-27
PTI-330: Pre-work: Reduce code duplication in epic import.

NickPhura:
This is some pre-IssuedTo code changes to make life easier, and reduce code volume in the epic integration.

### How it worked before
Each record type (order, inspection, etc) had its own class with its own complete set of transformation, document, save, etc, logic.

### How it works now
There is a base class, that contains the same code the previous record type utils did, with the exception that it only has the fields that all types have in common (which is almost all of them).
- It calls the pre-transform function
- It handles the document creation
- It transforms the common fields
- It contains the save logic

Each of the record type classes are now much simpler.  
- They extend the base class, and just append their unique fields to the output of the base classes transform method.

None of the actual logic has been changed, just re-organized to reduce duplication.


angular/package.json
```diff
@@ -41,8 +41,8 @@
     "lint": "npm-run-all -l -s -c lint:ts:* lint:html lint:scss",
     "lint:ts": "npm-run-all -l -s -c lint:ts:*",
     "lint:ts:1": "ng lint --format=stylish",
-    "lint:ts:2": "prettier projects/**/*.ts --warn",
-    "lint:html": "prettier projects/**/*.html --warn",
+    "lint:ts:2": "prettier projects/**/*.ts --check",
```
marklise: Lets not enable this right now.

api/src/integrations/epic/base-record-utils.js
```diff
@@ -0,0 +1,138 @@
...
+   * @memberof BaseRecordUtils
+   */
+  async saveDocument(epicRecord) {
+    const documents = [];
+
+    if (epicRecord && epicRecord._id && epicRecord.documentFileName) {
+      const savedDocument = await DocumentController.createDocument(
+        epicRecord.documentFileName,
+        (this.auth_payload && this.auth_payload.displayName) || '',
+        `https://projects.eao.gov.bc.ca/api/document/${epicRecord._id}/fetch/${encodeURIComponent(
```
marklise: Hostname of this part can be moved up to the top of the page as a const

angular/package.json
NickPhura: Enable it? It runs the same before and after. The only difference is that with `--check` it wont print the entire contents of files that fail.  Instead it will only print the name of the file.  Just makes it easier to read.

marklise: We ran into many issues with the environment vs. local, and it's separate from the bug that this task is supposed to be addressing.  We can file a separate PR instead.

api/src/integrations/epic/base-record-utils.js
NickPhura: I will fix this now

NickPhura: Updated

PR #211, opened 2020-03-05
Fix NRCED db-migrate script date format.

NickPhura:


NickPhura:
> ISO standard is dashes, not slashes.

That format is what moment is expecting to be passed, not what it outputs. 
- The CSV Kyle produced has dates in the format: `24/01/2019 0:00`.
The .toDate() call will still output a normal Date type.

marklise:
> @marklise
> 
> > ISO standard is dashes, not slashes.
> 
> That format is what moment is expecting to be passed, not what it outputs.
> 
> * The CSV Kyle produced has dates in the format: `24/01/2019 0:00`.
> 
> The .toDate() call should still output a normal Date type that satisfies mongoose/mongo.

Did the CSV change?

NickPhura:
> 
> 
> > @marklise
> > > ISO standard is dashes, not slashes.
> > 
> > 
> > That format is what moment is expecting to be passed, not what it outputs.
> > 
> > * The CSV Kyle produced has dates in the format: `24/01/2019 0:00`.
> > 
> > The .toDate() call should still output a normal Date type that satisfies mongoose/mongo.
> 
> Did the CSV change?

No?

Its whatever is in api/migrations/nrced-data.csv
- Just make sure to look at the raw csv, since excel applies date formatting by default.

PR #213, opened 2020-03-07
PTI-330: IssuedTo - Add admin/public add/edit/details controls.

NickPhura:


NickPhura:
> Names are being published on NRPTI Public when they do not have a DOB.
> Name changes in NRPTI admin are NOT being pushed to NRPTI public pages after a save:

There were some places in the controllers where for  whatever reason I was doing `incomgigObj.type`  when it should have been `incomgigObj.issuedTo.type` (etc).  So values weren't being set properly.

> I don't see the anonymous/!anonymous data filtering working quite right on the back-end. It's confusing on whether the anonymous field exists (as it's in the model, I'm not sure why you need this stored), vs. the public read role being on the object/sub object.

So currently, the anonymous field isn't really being used.  Its a feature for post-release, and its purpose is so that admins can manually set a record to anonymous, even if our business logic says its not anonymous.  What should be working as part of this PR, is just the read role being honoured.

There are also some hiccups due to the dirt-fields bug, which is why some of the anonymous field logic isn't functional as of yet.

The UI for it has also been hidden for now, so as not to confuse anyone when we release to prod.

marklise:
> So currently, the anonymous field isn't really being used. Its a feature for post-release, and its purpose is so that admins can manually set a record to anonymous, even if our business logic says its not anonymous. What should be working as part of this PR, is just the read role being honoured.
> 
> There are also some hiccups due to the dirty-fields bug, which is why some of the anonymous field logic isn't functional as of yet.
> 
> The UI for it has also been hidden for now, so as not to confuse anyone when we release to prod.

I don't believe in having the property 'anonymous' should be stored in the DB.  The role logic as a property of the object in question provides a higher level of granularity and flexibility than a simple property.  We don't need it.

NickPhura:
> 
> 
> > So currently, the anonymous field isn't really being used. Its a feature for post-release, and its purpose is so that admins can manually set a record to anonymous, even if our business logic says its not anonymous. What should be working as part of this PR, is just the read role being honoured.
> > There are also some hiccups due to the dirty-fields bug, which is why some of the anonymous field logic isn't functional as of yet.
> > The UI for it has also been hidden for now, so as not to confuse anyone when we release to prod.
> 
> I don't believe in having the property 'anonymous' should be stored in the DB. The role logic as a property of the object in question provides a higher level of granularity and flexibility than a simple property. We don't need it.

Yeah, you're right.

I've removed it from the models/controllers.

angular/projects/common/src/app/utils/utils.ts
```diff
@@ -38,6 +40,36 @@ export class Utils {
...
+
+    if (!obj.firstName && !obj.middleName && !obj.lastName) {
+      return '';
+    }
+
+    if (obj.type === ENTITY_TYPE.Individual) {
+      return [[obj.lastName || '-', obj.firstName || '-'].join(', '), obj.middleName || '-'].join(' ');
+    }
+  }
+
```
marklise: None of this belongs in a common utils class, it belongs to the Entity class itself as a member function.

angular/projects/common/src/app/utils/entity-type-enum.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @export
+ * @enum {number}
+ */
+export enum ENTITY_TYPE {
+  Company = 'Company',
+  Individual = 'Individual',
+  IndividualCombined = 'IndividualCombined',
+  NotSet = 'NotSet' // Not a supported type, used to indicate that an entity type has yet to be set.
+}
```
marklise: This belongs as part of the Entity class.

angular/projects/common/src/app/entity/entity-add-edit/entity-add-edit.component.html
```diff
@@ -0,0 +1,82 @@
...
+        <mat-slide-toggle
+          name="anonymous"
+          id="anonymous"
+          formControlName="anonymous"
+          (change)="updateAnonymousText($event.checked)"
+        >
+          {{ anonymousText }}
+        </mat-slide-toggle>
+      </div>
+    </ng-container> -->
```
marklise: No commented code.

angular/projects/admin-nrpti/src/assets/styles/components/add-edit.scss
```diff
@@ -127,7 +127,15 @@ legend {
   }
 }
 
+// Occupies the same space as a label-pair element, but contains only 1 element.
+.label-single {
+  @extend .label-pair;
+
+  padding-top: $fc-line-height;
+}
+
```
marklise: You're defining this, but then never using it?

angular/projects/common/src/app/entity/entity-detail/entity-detail.component.html
```diff
@@ -0,0 +1,59 @@
...
+    <span class="grid-item-name">Entity Type</span>
+    <span class="grid-item-value">{{ (data && data.type) || '-' }}</span>
+  </div>
+
+  <!--Individual* -->
+  <!-- TODO enable/update as part of anonymous entity/document work. -->
+  <!-- <div *ngIf="data && data.anonymous" class="grid-item">
+    <span class="grid-item-name">Anonymity</span>
+    <span class="grid-item-value">Personally identifiable information will not be published</span>
+  </div> -->
```
marklise: Remove

angular/projects/public-nrpti/src/app/records/administrative-sanctions/administrative-sanction-detail/administrative-sanction-detail.component.ts
```diff
@@ -16,10 +18,14 @@ export class AdministrativeSanctionDetailComponent implements OnInit, OnDestroy
   public loading = true;
   public activeTab = 'detail';
 
+  public entityString = '';
+
```
marklise: Once you put the function of building the string into a class, you won't need to declare this here anymore.

marklise:
Also, we don't need separate commits for lint, just do them as part of your code cleanup before submitting.  Please don't force push any edits to this PR any more as it breaks github tracking.

angular/projects/common/src/app/entity/entity-add-edit/entity-add-edit.component.html
NickPhura: So this is to support the anonymous stuff, which is up and coming 

> 
> 
> Also, we don't need separate commits for lint, just do them as part of your code cleanup before submitting. Please don't force push any edits to this PR any more as it breaks github tracking.

Some of the files linted arent related to this PR, but in theory would break the git actions that check linting.

I can put them in with the other commits, but then it would be misleading since no actual code was changed.

NickPhura:
> 
> 
> Also, we don't need separate commits for lint, just do them as part of your code cleanup before submitting. Please don't force push any edits to this PR any more as it breaks github tracking.

Some of the files linted arent related to this PR, but in theory would break the git actions that check linting.

I can put them in with the other commits, but then it would be misleading since no actual code was changed.

marklise:
> > Also, we don't need separate commits for lint, just do them as part of your code cleanup before submitting. Please don't force push any edits to this PR any more as it breaks github tracking.
> 
> Some of the files linted arent related to this PR, but in theory would break the git actions that check linting.
> 
> I can put them in with the other commits, but then it would be misleading since no actual code was changed.

I would have thought they broke the build before if that was the case.

NickPhura:
> 
> 
> > > Also, we don't need separate commits for lint, just do them as part of your code cleanup before submitting. Please don't force push any edits to this PR any more as it breaks github tracking.
> > 
> > 
> > Some of the files linted arent related to this PR, but in theory would break the git actions that check linting.
> > I can put them in with the other commits, but then it would be misleading since no actual code was changed.
> 
> I would have thought they broke the build before if that was the case.

Yes, that should be the case. But in this case they seem to have gotten past that check (maybe the lint checking isn't working properly).  Two of the files that lint-fix found, were the document-controller and nris datasource, neither of which I have changes for in this PR (or generally ever).

angular/projects/common/src/app/entity/entity-add-edit/entity-add-edit.component.html
NickPhura: This all has to do with the anonymous logic, which is related to issuedTo.

The only reason its commented out is because the dirty-fields issue breaks it, which I found out after writing all of it.

So I think it should stay, and either get un-commented when the dirty-fields PRs are all done OR this whole PR can wait for them to get merged, and I can un-comment it all at once as the last step of this PR.

angular/projects/public-nrpti/src/app/records/administrative-sanctions/administrative-sanction-detail/administrative-sanction-detail.component.ts
NickPhura: As far as I can tell this is still necessary (and accompanying logic).

I tried to move this (`new Entity(this.data.issuedTo).buildEntityString()`) to the HTML, but angular doesn't like it. 
I did try declaring the class in the TS, so that the HTML would recognize `Entity` but no dice.

marklise:
Broken Admin. See screenshots:

Current PR:
Screen Shot 2020-03-27 at 3 39 09 PM

Dev:
Screen Shot 2020-03-27 at 3 40 35 PM

This doesn't look like what the JIRA task is asking for.
Individual:
image
Company:
image

Please go through all the areas you've touched and do a regression test.

NickPhura:
> 
> 
> Broken Admin. See screenshots:
> 
> Current PR:
> Screen Shot 2020-03-27 at 3 39 09 PM
> 
> Dev:
> Screen Shot 2020-03-27 at 3 40 35 PM
> 
> This doesn't look like what the JIRA task is asking for.
> Individual:
> image
> Company:
> image
> 
> Please go through all the areas you've touched and do a regression test.

Hard to say exactly what you're referring to.

The one thing I notice is that the Company Name input didnt open, even tho the Entity Type is selected.

I pushed a commit to fix that.  When I moved some of the old code into the Entity.ts class, that import didnt get updated.  Not sure why my local build didnt get angry :/

api/src/utils/query-actions.js
```diff
@@ -21,7 +21,7 @@ exports.publish = async function(obj) {
       });
     } else {
       // publish
-      obj.read.push('public');
+      self.addPublicReadRole(obj);
       obj.markModified('read');
```
marklise, line 25: This should probably go inside the addPublicReadRole function

api/src/utils/query-actions.js
```diff
@@ -66,7 +82,7 @@ exports.unPublish = async function(obj) {
       });
     } else {
       // unpublish
-      obj.read = obj.read.filter(role => role !== 'public');
+      self.removePublicReadRole(obj);
       obj.markModified('read');
```
marklise, line 86: Same here a/a

marklise:
Date of birth changes are not making the records anonymous on the public side when they are moved into an age limit that should make the record info anonymous.

PR #218, opened 2020-03-09
PTI-257: Add/edit documents for all record types.

danieltruong:


NickPhura:
Looks like there are some front/back end linting issues. 

The git checks say they passed, but looking at their output, it does show lint issues.  Something must be not quite right with either the git check or the lint command, since lint errors aren't actually failing the check. That's a separate issue to fix though.

NickPhura:
One thing I noticed is that when you enter a Document URL, if you don't specify `https://` it will try to open the href relative to the current domain.

For example:  
- I put in `www.google.com` but it tries to open `http://localhost:4200/www.google.com`
- If I put `https://www.google.com` then it works as expected
  - `//www.google.com` also works, but thats more of a fun-fact.

angular/projects/admin-nrpti/src/app/records/utils/record-utils.ts
```diff
@@ -230,4 +230,46 @@ export class RecordUtils {
         return null;
     }
   }
+
+  async handleDocumentChanges(links = [], documents = [], documentsToDelete = [], recordId, factoryService) {
```
NickPhura: Id recommend adding some types to these arrays. Ex: `links = string[]` etc

I would also be my preference to throw some JSDoc on this function, since its not particularly clear what these arrays hold.  Ex: Does `documentsToDelete` have whole document objects in it? or Just document IDs?

angular/projects/admin-nrpti/src/app/records/utils/record-utils.ts
```diff
@@ -230,4 +230,46 @@ export class RecordUtils {
...
+
+    // Handle adding links
+    links.forEach(async link => {
+      promises.push(factoryService.createDocument(link, recordId));
+    });
+
+    // TODO: Handle adding S3 Docs
+    console.log(documents);
+
+    // TODO: Handle deleting documents
```
NickPhura: Are these still active TODO's? Or can these comments be removed?

danieltruong:
> One thing I noticed is that when you enter a Document URL, if you don't specify `https://` it will try to open the href relative to the current domain.
> 
> For example:
> 
> * I put in `www.google.com` but it tries to open `http://localhost:4200/www.google.com`
> * If I put `https://www.google.com` then it works as expected
>   
>   * `//www.google.com` also works, but thats more of a fun-fact.
> 
> So we probably need to check/add `https://` to any urls before saving, otherwise lots of links will probably end up broken. Or whatever elegant way there is to handle this, maybe its a common piece of code somewhere we can steal. Possibly running it through `new URL()` might help?

I added new URL() as a check. Works well and also automatically changes http to https.

angular/projects/common/src/app/document-link-staging/document-link-staging.component.ts
```diff
@@ -52,7 +52,13 @@ export class DocumentLinkStagingComponent {
...
       }
       const link = new Document();
-      link.url = this.linkUrl;
+      try {
+        link.url = new URL(this.linkUrl).href;
+      } catch (e) {
+        alert('You must submit a valid url.');
+        this.linkUrl = '';
+        return;
+      }
```
NickPhura: I don't think this quite covers the issue with the links.

The user can't put a bad url in, which is good, but it doesn't tell them what a valid url is, and more ideally, doesn't just add the 'https://' for them automatically.

I tried it out, and this npm package does what we need, and is apparently quite popular: https://www.npmjs.com/package/normalize-url

We can still use `new URL()` to validate that the url is ultimately valid, but the normalize call will at least add `https://` for the user, and will even change `http://` to `https://` as well.

```
      try {
        link.url = normalizeUrl(this.linkUrl, {
          forceHttps: true,
          stripWWW: false,
          stripAuthentication: false,
          removeTrailingSlash: false,
          sortQueryParameters: false
        });

        link.url = new URL(link.url).href;
      } catch (e) {
        alert('You must submit a valid url.');
        this.linkUrl = '';
        return;
      }
```

One important note: would need to use `normalize-url@4.5.0` rather than `5.0.0`, as they use non-backwards compatible regex's in 5.0.0 that don't transpile down to es5 and so break our app.

danieltruong: I don't think we need to enforce https. The user can copy and paste a url in. new URL will check if the url is http or https to prevent https://www.nrpti.gov.bc.ca/www.google.com from happening. If a user decides to copy and paste an invalid url while uploading a document, so be it. They'll likely notice when the link doesn't work and re-upload accordingly.

danieltruong: User may also need to link to some docs with only http. By forcing a url with http to https, it might prevent them from linking.

NickPhura: In regards to invalid urls.  The current code wont work if they paste 'www.google.com'.  It will say its invalid.  That was the main part I thought should be adjusted.  Since most average users won't know why that is not valid, or what the http(s):// part even is.  If we can auto-add it, when its missing, then it will make things much simpler for them.

In regards to http links: I though there were some rules about having http links in an https site? I cant remember exactly but I thought an https site wasn't allowed to link out to non https sites, or something like that?

marklise: Pasted URLs from chrome, while viewing them in the browser they look like they have the protocol part missing, will in fact get inserted into the clipboard upon copy - so no issue on copy/paste there.  (Debate over chrome removing this key part of the address is interesting to say the least)
Firefox/IE prefix the protocol display all the time, so also non-issue there.
Copy hyperlink from MS Word, prefixes the protocol (not true if you select the text manually - however, I don't believe we should be encouraging this type of behaviour).

The RFC shows that absent 'http' or 'https', it is violating the concept of a link and should not be considered valid.

I don't believe in forcing the protocol either, because wherever the link they are copying from may in fact be on an insecure page, where they (the server) do not support https protocol.  Forcing https on a server that doesn't serve out content via https will create broken links, and only visible upon the user/public clicking on said link and the server in question delivering anything.  Not really our problem as the web is full of insecure content.  So, I think the user should be told to provide a valid link which must include the protocol.

Your comment around http/https in a page rendered by the browser relates to the content and the source of said content, not the meta detailing anchor tags/hrefs.  Mixed content is a bad thing and lowers the security of the entire page, but that is not applied to the meta that was already served over https (in this case, the page rendering the html with an anchor tag).  If there was content in the page that referenced an image, or iframe, or video... etc that was itself being served over http in an https page, then that would throw the error you are talking about.

NickPhura: Sure, at the very least then, the error message should detail what a valid link is.  Mentioning the `http://` or `https://` bit, so they know what the fix if they find themselves trying to enter a manually copied text link, which is for sure going to happen at some point.

marklise: Yep exactly - that's what I meant by telling the user to provide a valid link.  Up to you two on how you want to achieve that.

danieltruong: Updated to alert('You must submit a valid url. Be sure to include http:// or https:// as part of your URL.');

angular/projects/admin-nrpti/src/app/services/record.service.ts
```diff
@@ -46,12 +46,24 @@ export class RecordService {
...
   }
 
-  createRecord(record: object): Observable<object> {
+  createRecord(record: any): Observable<object> {
+    // We handle document logic when we add or remove documents not when we add/edit record
+    for (const property of Object.keys(record)) {
+      record[property].forEach(element => {
+        delete element.documents;
+      });
+    }
```
marklise: We shouldn't have to iterate through all the documents on the client side if the API just ignores it.

angular/projects/admin-nrpti/src/app/services/record.service.ts
```diff
@@ -46,12 +46,24 @@ export class RecordService {
...
   }
 
-  editRecord(record: object): Observable<object> {
+  editRecord(record: any): Observable<object> {
+    // We handle document logic when we add or remove documents not when we add/edit record
+    for (const property of Object.keys(record)) {
+      record[property].forEach(element => {
+        delete element.documents;
+      });
+    }
```
marklise: A/A re: iterating through documents.

danieltruong: Fixed in c21738f

angular/projects/admin-nrpti/src/app/services/record.service.ts
danieltruong: Fixed in c21738f

api/src/controllers/record-controller.js
```diff
@@ -44,7 +44,7 @@ let EditManagementPlan = require('./put/management-plan');
  * @param {*} res
  * @param {*} next
  */
-exports.protectedOptions = function(args, res, next) {
+exports.protectedOptions = function (args, res, next) {
```
NickPhura: I noticed elsewhere that the old lint issue of `{}` getting changed to `{ }` was fixed, so thats good.

But looks like there is still one last setting in your local vscode that is adding spaces between `function` and the brackets.

NickPhura: Its this setting, I believe: `insertSpaceBeforeFunctionParenthesis`

It defaults to false, so somewhere it must have gotten toggled to true.

danieltruong: I run lint-fix before every commit. This has nothing to do with my VS Code settings. So lint-fix is not catching this.

NickPhura: When I run lint-fix locally, it removes those spaces.  The only reason it wouldn't, is if the config file was overwritten by another setting.

PR #273, opened 2020-04-01
PTI-486: Removed tooltip from document links

danieltruong:


danieltruong:
> Fix `target="_blank"`

resolved in e4028ba

PR #277, opened 2020-04-03
PTI-470: Improve epic import to support updates + new arch/models.

NickPhura:
PTI-408: Refine import to only fetch some record types from LNG projects only.

## Updates:
- Supports the new arch/models + supports updates properly
  - Tries to find an existing record
    - If Found
      - Transform record and call recordController.processPutRequest which has all of the logic for editing records, including their flavours
    - If Not Found
      - Transform record and call recordController.processPostRequest which has all of the logic for creating records, including their flavours
- Utilizes the taskAuditRecord to update counts/totals live.

### DIFF
--- a/api/src/integrations/epic/base-record-utils.js
+++ b/api/src/integrations/epic/base-record-utils.js
@@ -76,6 +76,16 @@ class BaseRecordUtils {
 
     const documentIds = nrptiRecord.documents.map(id => new ObjectID(id));
 
+    // Check if any documents uploaded to S3, and delete if any
+    for (let idx = 0; idx < documentIds.length; idx++) {
+      const document = await DocumentModel.findOne({ _id: documentIds[idx] });
+
+      if (document && document.key) {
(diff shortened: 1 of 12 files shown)


api/src/integrations/epic/base-record-utils.js
```diff
@@ -58,6 +60,25 @@ class BaseRecordUtils {
...
+      return;
+    }
+
+    const DocumentModel = mongoose.model('Document');
+
+    const documentIds = nrptiRecord.documents.map(id => new ObjectID(id));
+
+    await DocumentModel.deleteMany({ _id: { $in: documentIds } });
+  }
+
```
marklise, line 81: This just deletes the document meta, what about the actual files if we stored them locally/s3 etc?

NickPhura: Fair point.  I'm updating it currently so that It fetches the documents, checks if an s3 key is present, and then deletes from s3 in addition to deleting locally.

PR #284, opened 2020-04-03
Remove 'Warning' from NRCED activity type filters

NickPhura:
Remove 'Warning' from NRCED activity type filters
Remove 'ALL' option from records-list page size
Update issuedTo display string to not show dashes, etc.
Show 'Unpublished' on NRCED if no issuedTO object exists (redacted).
Update nrced migration to not add issuedTo public role if its an anonymous person.

### DIFF
--- a/angular/projects/common/src/app/utils/record-constants.ts
+++ b/angular/projects/common/src/app/utils/record-constants.ts
@@ -48,8 +48,7 @@ export class Picklists {
     Inspection: { displayName: 'Inspection', _schemaName: 'InspectionNRCED' },
     Order: { displayName: 'Order', _schemaName: 'OrderNRCED' },
     RestorativeJustice: { displayName: 'Restorative Justice', _schemaName: 'RestorativeJusticeNRCED' },
-    Ticket: { displayName: 'Ticket', _schemaName: 'TicketNRCED' },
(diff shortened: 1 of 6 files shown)


angular/projects/common/src/app/utils/record-constants.ts
```diff
@@ -48,8 +48,7 @@ export class Picklists {
     Inspection: { displayName: 'Inspection', _schemaName: 'InspectionNRCED' },
     Order: { displayName: 'Order', _schemaName: 'OrderNRCED' },
     RestorativeJustice: { displayName: 'Restorative Justice', _schemaName: 'RestorativeJusticeNRCED' },
-    Ticket: { displayName: 'Ticket', _schemaName: 'TicketNRCED' },
-    Warning: { displayName: 'Warning', _schemaName: 'WarningNRCED' }
+    Ticket: { displayName: 'Ticket', _schemaName: 'TicketNRCED' }
```
marklise, line 51: Are we supposed to remove Warning?  I didn't see it mentioned in the commit msg.

NickPhura: Yeah, apparently there is some fuzziness at the business level around whether or not Warnings should be on NRCED, and possibly if they are allowed to be even (due to some law/rule or something).  Lena was fine with removing it, for now, to appease the business.

PR #294, opened 2020-04-07
PTI-509: NRCED Homepage tweaks.

marklise:


NickPhura:
Its not super clear in the ticket, but they wanted a `Find Records` button in the home page banner:

image

NickPhura:
Ugh, I really hate the `close and comment` button...

marklise:
I think we need a quick mock up to be clear then.  Maybe we can land this and address the final nit tomorrow.

PR #304, opened 2020-04-08
Remove populate from NRCED list resolver + UI Tweaks

NickPhura:
- Records List Resolver - no longer populate documents
- Record List Detail components - fetch documents on init (on expand)
- UI - Fix vertical alignment of table values when collapsed


angular/projects/public-nrpti/src/app/records/administrative-penalties/administrative-penalty-detail/administrative-penalty-detail.component.ts
```diff
@@ -17,12 +19,20 @@ export class AdministrativePenaltyDetailComponent implements OnInit, OnDestroy {
...
+    public _changeDetectionRef: ChangeDetectorRef
+  ) {}
 
   ngOnInit(): void {
     if (this.data) {
       this.data = new AdministrativePenaltyNRCED(this.data);
 
+      // populate documents
+      this.getDocuments();
+
```
marklise, line 35: This call is useful (see second note below for follow up)

angular/projects/public-nrpti/src/app/records/administrative-penalties/administrative-penalty-detail/administrative-penalty-detail.component.ts
```diff
@@ -39,11 +49,40 @@ export class AdministrativePenaltyDetailComponent implements OnInit, OnDestroy {
       this.data =
         (res.records[0] && res.records[0].data && new AdministrativePenaltyNRCED(res.records[0].data)) || null;
 
+      // populate documents
+      this.getDocuments();
+
```
marklise, line 54: This call is not necessary.  In fact, after reviewing all the other pages components, the resolvers aren't needed as we have no routes that need resolving.  So it's currently not necessary to include this function call here, the AdministrativePenaltyResolver, or this particular snip of code, which is copy/pasted on all other detail components.

```this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
      if (!res || !res.records) {
        alert("Uh-oh, couldn't load administrative penalty");
        this.router.navigate(['/']);
        return;
      }

      // If data was passed in directly, take it over anything in the route resolver.
      this.data =
        (res.records[0] && res.records[0].data && new AdministrativePenaltyNRCED(res.records[0].data)) || null;

      // populate documents
      this.getDocuments();

      this.loading = false;
      this._changeDetectionRef.detectChanges();
    });

NickPhura: So I started to strip this stuff out, but tested to see if it all works first (the direct link to a record) since its been a while since it was written, and it does.  So rather that purge all of this work, I added a button to copy a link to the record.  

It was something we talked about adding before, and the logic was pretty straight forward, so seemed worthwhile to do.

NickPhura:
> 
> 
> This commit #eaa8470 should probably go in a separate PR. The issues above still need to be resolved (clean out the unused code).

The router code is no longer unused, now that the copy link other code as been added.  If you copy the link, and navigate to it, the component loads using that router logic.

marklise:
Closing due to changes in repo

PR #318, opened 2020-04-16
PTI-536: Anonymous Records

NickPhura:


api/src/controllers/document-controller.js
```diff
@@ -69,14 +79,7 @@ exports.protectedPost = async function (args, res, next) {
...
-          .upload({
-            Bucket: process.env.OBJECT_STORE_bucket_name,
-            Key: docResponse.key,
-            Body: args.swagger.params.upfile.value.buffer,
-            ACL: 'public-read'
-          })
-          .promise();
+        const s3UploadResult = await uploadS3Document(docResponse.key, args.swagger.params.upfile.value.buffer);
 
         s3Response = s3UploadResult;
```
marklise, line 84: I think we can save ourselves a variable initialization here.  Just assign s3Response to the await uploadS3Document

api/src/controllers/document-controller.js
```diff
@@ -254,13 +263,153 @@ async function createDocument(fileName, addedBy, url = null) {
...
 exports.deleteS3Document = deleteS3Document;
 
 async function deleteS3Document(docKey) {
   return await s3.deleteObject({ Bucket: process.env.OBJECT_STORE_bucket_name, Key: docKey }).promise();
+}
+
+exports.uploadS3Document = uploadS3Document;
+
+async function uploadS3Document(docKey, docBody) {
+  // TODO: ACL is set to public-read until we know publishing rules.
```
marklise, line 285: Is this a real TODO?

api/src/controllers/document-controller.js
```diff
@@ -254,13 +263,153 @@ async function createDocument(fileName, addedBy, url = null) {
...
+ * @param {*} res
+ * @param {*} next
+ */
+exports.protectedPublish = async function(args, res, next) {
+  if (!args.swagger.params.recordId || !args.swagger.params.recordId.value) {
+    return queryActions.sendResponse(res, 400, 'MIssing required recordId param');
+  }
+
+  if (!args.swagger.params.documentId || !args.swagger.params.documentId.value) {
+    return queryActions.sendResponse(res, 400, 'MIssing required documentId param');
```
marklise, line 309: *MIssing => Missing

api/src/controllers/document-controller.js
```diff
@@ -254,13 +263,153 @@ async function createDocument(fileName, addedBy, url = null) {
...
+    defaultLog.info(`publishDocument - couldn't find document for documentId: ${documentId}`);
+    throw {
+      name: 'MissingDocumentError',
+      message: `publishDocument - couldn't find document for documentId: ${documentId}`
+    };
+  }
+
+  const db = mongodb.connection.db(process.env.MONGODB_DATABASE || 'nrpti-dev');
+  const collection = db.collection('nrpti');
+  const masterRecord = await collection.findOne({ _id: new ObjectID(recordId) });
```
marklise, line 346: All we need is to require mongoose, we don't need a new db connection.

api/src/controllers/document-controller.js
```diff
@@ -254,13 +263,153 @@ async function createDocument(fileName, addedBy, url = null) {
...
   document.url = url;
   document.key = `${document._id}/${fileName}`;
-  document.read = ['public', 'sysadmin'];
+  document.read = ['sysadmin'];
   document.write = ['sysadmin'];
+
+  if (!isAnonymous) {
+    document.read.push('public');
+  }
+
```
marklise, line 272: I don't think whether the individual/company is anonymous or not should determine the document publishing rule.  It should be separate from the anonymous logic, and controlled via a different flag.  Maybe just the terminology is wrong, but the roles to be added should probably be passed into the function, as opposed to a simple binary flag.

api/src/controllers/put/administrative-penalty.js
```diff
@@ -110,19 +111,48 @@ exports.editRecord = async function(args, res, next, incomingObj) {
...
+
+    if (isAnonymous) {
+      savedAdministrativePenalty.documents.forEach(docId => {
+        documentPromises.push(DocumentController.unpublishDocument(docId, args.swagger.params.auth_payload));
+      });
+    } else if (savedAdministrativePenalty) {
+      savedAdministrativePenalty.documents.forEach(docId => {
+        DocumentController.publishDocument(savedAdministrativePenalty._id, docId, args.swagger.params.auth_payload);
+      });
+    }
```
marklise, line 137: Since you're not using isAnonymous anywhere else, I'd rather save some vertical space and use:
`if (QueryUtils.isRecordAnonymous(savedAdministrativePenalty)) {`

There's a few of these in the PR.

api/src/controllers/put/administrative-sanction.js
```diff
@@ -35,6 +36,7 @@ const AdministrativeSanctionPost = require('../post/administrative-sanction');
  * @returns object containing the operation's status and created records
  */
 exports.editRecord = async function(args, res, next, incomingObj) {
+  console.log('1', incomingObj);
```
marklise, line 39: Remove

api/src/controllers/put/administrative-sanction.js
```diff
@@ -306,6 +337,7 @@ exports.editLNG = async function(args, res, next, incomingObj) {
  * @returns edited nrced administrativeSanction record
  */
 exports.editNRCED = async function(args, res, next, incomingObj) {
+  console.log('2', incomingObj);
```
marklise, line 340: Remove.

api/src/utils/query-utils.js
```diff
@@ -121,6 +121,15 @@ exports.isRecordAnonymous = function(record) {
     return false;
   }
 
+  if (record.issuedTo.forceAnonymous) {
+    // record manually set to anonymous
+
+    // Note: this value only indicates that the user has toggled this record to be anonymous.  If it is set to null or
+    //   false, that only tells us the user hasn't toggled this value, and does not necessarily mean this record is
+    //   public (not anonymous).
```
marklise, line 129: Alignment of comment text.

Additionally I think the function name should be something like "isConsideredAnonymous", because we may want to have different rules around what defines anonymous based on particular record types.  It's possible LNG, NRCED, and the incoming 3rd flavour may want to treat things a bit differently - either by record type, or by business area.

api/src/utils/query-actions.js
```diff
@@ -28,8 +28,8 @@ exports.publish = async function(obj) {
       obj.datePublished = date;
       obj.markModified('datePublished');
 
-      obj.dateUpdated = date;
-      obj.markModified('dateUpdated');
```
marklise, line 32: I think we want to keep this level of detail.  Same for unPublish

api/src/utils/query-actions.js
```diff
@@ -28,8 +28,8 @@ exports.publish = async function(obj) {
...
+      obj.publishedBy = auth_payload && auth_payload.displayName;
```
marklise, line 31: I don't think we want display name here.  We should be using a unique identifier here for the user: idir_userid.  We may want to start storing a complex object here instead of a string.  ie: the preferred_username & idir_userid.  I don't think we should store displayName at all.

danieltruong:
> Quick overview of the changes:
> 
> `forceAnonymous` is a new field in the `issuedTo` object.
> 
> * It indicates if the user has manually enforced anonymity or not.
> * It is not used to determine redactions, etc, that is all done through the read roles still
> 
> Documents now have their read roles updated based on the anonymity checks against a records `issuedTo` object.
> 
> * This is only checked for record types whose schema has `issuedTo`.

Why would we need a forceAnonymous. From what I understand, this flag is able to hide a users information on selected front ends if it is set as true. For example, on NRCED the issued to will not reveal the persons details, however, if we look in NRPTI, the persons details will remain. Why not use the existing roles and permissions paradigm we have and just tag read as [sysadmin] or [public, sysadmin] instead. That way there isn't another thing for us to manage when displaying data in our front ends or future front ends. It also allows us to be able to control according to types of users. So say, there's something we want to hide from a BCMI Admin, it can be easily done with very little code.

Looking at our current code in master, it looks like issuedTo already has read, write fields.

api/src/controllers/put/administrative-sanction.js
NickPhura: > 
> 
> > Quick overview of the changes:
> > `forceAnonymous` is a new field in the `issuedTo` object.
> > 
> > * It indicates if the user has manually enforced anonymity or not.
> > * It is not used to determine redactions, etc, that is all done through the read roles still
> > 
> > Documents now have their read roles updated based on the anonymity checks against a records `issuedTo` object.
> > 
> > * This is only checked for record types whose schema has `issuedTo`.
> 
> Why would we need a forceAnonymous. From what I understand, this flag is able to hide a users information on selected front ends if it is set as true. For example, on NRCED the issued to will not reveal the persons details, however, if we look in NRPTI, the persons details will remain. Why not use the existing roles and permissions paradigm we have and just tag read as [sysadmin] or [public, sysadmin] instead. That way there isn't another thing for us to manage when displaying data in our front ends or future front ends. It also allows us to be able to control according to types of users. So say, there's something we want to hide from a BCMI Admin, it can be easily done with very little code.
> 
> Looking at our current code in master, it looks like issuedTo already has read, write fields.



> 
> 
> > Quick overview of the changes:
> > `forceAnonymous` is a new field in the `issuedTo` object.
> > 
> > * It indicates if the user has manually enforced anonymity or not.
> > * It is not used to determine redactions, etc, that is all done through the read roles still
> > 
> > Documents now have their read roles updated based on the anonymity checks against a records `issuedTo` object.
> > 
> > * This is only checked for record types whose schema has `issuedTo`.
> 
> Why would we need a forceAnonymous. From what I understand, this flag is able to hide a users information on selected front ends if it is set as true. For example, on NRCED the issued to will not reveal the persons details, however, if we look in NRPTI, the persons details will remain. Why not use the existing roles and permissions paradigm we have and just tag read as [sysadmin] or [public, sysadmin] instead. That way there isn't another thing for us to manage when displaying data in our front ends or future front ends. It also allows us to be able to control according to types of users. So say, there's something we want to hide from a BCMI Admin, it can be easily done with very little code.
> 
> Looking at our current code in master, it looks like issuedTo already has read, write fields.

The `forcedAnonymous` flag is there because an admin user can choose to make a record anonymous, even if they have a birth date and are an adult.  The reason we need it, is because we need to distinguish between records that are anonymous because of the underage rules, and those that are anonymous because we've manually set them to be.

If you tried to update a record that had an adult person (valid birth date, etc), but they were manually set to anonymous, how would you know NOT to update the read roles?  Because the business logic says they should be public, but the admin wants them to remain anonymous.  So we have to store that value.  We still use the read roles to determine redactions like normal, that `forceAnonymous` flag is only there to indicate that the admin wants to disregard the business rules around anonymity, and force it to be anonymous (presumably because they have a good reason).

api/src/controllers/put/administrative-sanction.js
NickPhura: Bleh, Fill fix, and the other console logs

api/src/utils/query-utils.js
NickPhura: Yeah I intentionally indented the comment a bit so it would stand out as a paragraph, rather than separate lines. But I'm not worried about the format, so I can change it.

Yeah I can change the function name, sounds good.

api/src/utils/query-actions.js
NickPhura: My thoughts at the time, were around whether or not we consider toggling the publish state as an update/edit or not.  But I could see the argument either way.  Can put it back.

api/src/utils/query-actions.js
NickPhura: TBH Im not sure what values come in the auth_payload.  We seem to use `displayName` in the vast majority of places, so I just followed suit.  But, can change it to whatever is most appropriate of course.

api/src/controllers/document-controller.js
NickPhura: Not sure, would have to ask @danieltruong 

I only moved this to a function, but didn't touch any of the S3 related logic (I don't have any experience with S3, or any of nrpti's S3 code).

api/src/controllers/document-controller.js
```diff
@@ -15,33 +16,40 @@ const s3 = new AWS.S3({
...
+    // Determine if the record is anonymous, and therefore the document must not be publicly available.
+    let isAnonymous = false;
+    const db = mongodb.connection.db(process.env.MONGODB_DATABASE || 'nrpti-dev');
+    const collection = db.collection('nrpti');
+    const masterRecord = await collection.findOne({ _id: new ObjectID(args.swagger.params.recordId.value) });
+    if (Object.prototype.hasOwnProperty.call(mongoose.model(masterRecord._schemaName).schema.obj, 'issuedTo')) {
+      // Only check for anonymity if the master schema has entity information.  Otherwise assume not anonymous.
+      isAnonymous = queryUtils.isRecordAnonymous(masterRecord);
+    }
+
```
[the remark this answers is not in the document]
NickPhura: It was using a db connection like this before, I just moved it to the start of the function since I added another place it gets used down below.  But, I think you're right, so Ill update it here and below.

NickPhura:
> 
> 
> > Quick overview of the changes:
> > `forceAnonymous` is a new field in the `issuedTo` object.
> > 
> > * It indicates if the user has manually enforced anonymity or not.
> > * It is not used to determine redactions, etc, that is all done through the read roles still
> > 
> > Documents now have their read roles updated based on the anonymity checks against a records `issuedTo` object.
> > 
> > * This is only checked for record types whose schema has `issuedTo`.
> 
> Why would we need a forceAnonymous. From what I understand, this flag is able to hide a users information on selected front ends if it is set as true. For example, on NRCED the issued to will not reveal the persons details, however, if we look in NRPTI, the persons details will remain. Why not use the existing roles and permissions paradigm we have and just tag read as [sysadmin] or [public, sysadmin] instead. That way there isn't another thing for us to manage when displaying data in our front ends or future front ends. It also allows us to be able to control according to types of users. So say, there's something we want to hide from a BCMI Admin, it can be easily done with very little code.
> 
> Looking at our current code in master, it looks like issuedTo already has read, write fields.

The `forcedAnonymous` flag is there because an admin user can choose to make a record anonymous, even if they have a birth date and are an adult.  The reason we need it, is because we need to distinguish between records that are anonymous because of the underage rules, and those that are anonymous because we've manually set them to be.

If you tried to update a record that had an adult person (valid birth date, etc), but they were manually set to anonymous, how would you know NOT to update the read roles?  Because the business logic says they should be public, but the admin wants them to remain anonymous.  So we have to store that value.  We still use the read roles to determine redactions like normal, that `forceAnonymous` flag is only there to indicate that the admin wants to disregard the business rules around anonymity, and force it to be anonymous (presumably because they have a good reason).

api/src/controllers/put/administrative-penalty.js
NickPhura: This looks a bit weird, possibly I did a find/replace when I made some updates, and it was left looking a bit odd.  I can update this.

danieltruong:
> > > Quick overview of the changes:
> > > `forceAnonymous` is a new field in the `issuedTo` object.
> > > 
> > > * It indicates if the user has manually enforced anonymity or not.
> > > * It is not used to determine redactions, etc, that is all done through the read roles still
> > > 
> > > Documents now have their read roles updated based on the anonymity checks against a records `issuedTo` object.
> > > 
> > > * This is only checked for record types whose schema has `issuedTo`.
> > 
> > 
> > Why would we need a forceAnonymous. From what I understand, this flag is able to hide a users information on selected front ends if it is set as true. For example, on NRCED the issued to will not reveal the persons details, however, if we look in NRPTI, the persons details will remain. Why not use the existing roles and permissions paradigm we have and just tag read as [sysadmin] or [public, sysadmin] instead. That way there isn't another thing for us to manage when displaying data in our front ends or future front ends. It also allows us to be able to control according to types of users. So say, there's something we want to hide from a BCMI Admin, it can be easily done with very little code.
> > Looking at our current code in master, it looks like issuedTo already has read, write fields.
> 
> The `forcedAnonymous` flag is there because an admin user can choose to make a record anonymous, even if they have a birth date and are an adult. The reason we need it, is because we need to distinguish between records that are anonymous because of the underage rules, and those that are anonymous because we've manually set them to be.
> 
> If you tried to update a record that had an adult person (valid birth date, etc), but they were manually set to anonymous, how would you know NOT to update the read roles? Because the business logic says they should be public, but the admin wants them to remain anonymous. So we have to store that value. We still use the read roles to determine redactions like normal, that `forceAnonymous` flag is only there to indicate that the admin wants to disregard the business rules around anonymity, and force it to be anonymous (presumably because they have a good reason).

Are there two different types of anonymous that we need to preserver? Because regardless of why the information is anonymous the action of making something anonymous is the same. It's information we don't want to surface.

We should maintain role based access control. We can keep all the business logic. Just map it to how we already handle permissions. So in add edit, we can have a checkbox or whatever that forces anonymous. In our api code, we would remove public from read specifically for issued to for the given record.

api/src/controllers/document-controller.js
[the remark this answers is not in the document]
NickPhura: Actually, I don't think we can change this to just mongoose.

We only know the incoming recordId, we don't know what the model is.  And mongoose always requires that you pick a model, I believe?

PR #328, opened 2020-04-20
PTI-513: Adjust public nrced date range filter UI.

NickPhura:
- Now has a single header with date format, instead of one per date field.
- UI only, no logic updates.

Addresses the last bit of PTI-513 (that is not struck through).


NickPhura:
> 
> 
> https://stackoverflow.com/questions/12894169/what-is-the-html-for-attribute-in-label
> 
> If anything, the for= should be an English word or phrase.

I wasn't quite sure what the best way to use `for` was in this case.

We don't really have (label, input) pairs, since all of the inputs are components. And in some cases, we have 1 label for 2 components.

I could add `id` to each of the components, and then add a matching `for` to the label.  Apparently you can have a for with multiple `id`, IE: `for="id1 id2"`

Thoughts?

angular/projects/public-nrpti/src/app/records/records-list/search-filters/search-filters.component.html
```diff
@@ -18,16 +18,21 @@ <h3>Filter</h3>
...
+          </div>
         </div>
-        <span class="date-separator align-self-center">To</span>
-        <div class="label-pair">
+        <!-- <div class="label-pair">
           <label for="recordType">Issued Date</label>
+          <span class="grey-subtext-sm ml-2 ">(YYYY-MM-DD)</span>
           <lib-date-picker [control]="formGroup.get('dateIssuedEnd')" [reset]="resetControls"></lib-date-picker>
-        </div>
+        </div> -->
```
marklise: Remove the code entirely

marklise:
> I could add `id` to each of the component tags, and then add a matching `for` to the label

This - we should probably assign unique id and link it to the for.

PR #337, opened 2020-04-27
PTI-564: Move RecordName field from master/nrced/lng to lng

NickPhura:
- update controllers
  - removed recordName from the master and nrced sections
- update add-edit components
  - moved recordName from root level of the form/save obj, and moved it to the lng section/sub-object
- update details components
  - removed recordName from master details component, and moved it to the lng flavour detail component
- update models
  - removed recordName from master and nrced models, for both api and angular models
- update migrations
  - removed recordName completely from nrced mirgration
  - removed recordName from master record in lng migration

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/administrative-penalties/administrative-penalty-add-edit/administrative-penalty-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/records/administrative-penalties/administrative-penalty-add-edit/administrative-penalty-add-edit.component.ts
@@ -101,7 +101,6 @@ export class AdministrativePenaltyAddEditComponent implements OnInit, OnDestroy 
   private buildForm() {
     this.myForm = new FormGroup({
       // Master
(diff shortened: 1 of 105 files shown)


danieltruong:
This ticket was a deferred ticket. To resolve this, I think we should keep recordName as part of master. This is why add/edit still has recordName as part of the master even though visually the form field for record name is in the LNG section. Record name is a field that will likely be useful when other public sites land such as BCMI. To resolve PTI-564, we should consider only doing a change in the UI and no where else.

NickPhura:
> 
> 
> This ticket was a deferred ticket. To resolve this, I think we should keep recordName as part of master. This is why add/edit still has recordName as part of the master even though visually the form field for record name is in the LNG section. Record name is a field that will likely be useful when other public sites land such as BCMI. To resolve PTI-564, we should consider only doing a change in the UI and no where else.

I think you're right.  I'm going to close this PR, and Ill open another that just makes that UI change.

PR #346, opened 2020-05-01
PTI-536: Anonymous Records

NickPhura:
### Commit 1 - API Changes
- Updates to document controller to support public/non-public roles and signed urls.
- Updates to put/post controllers for the 8 C&E record types, to account for anonymous isssuedTo/Document business logic.

### Commit 2 - Admin Changes
- This is mostly just the code needed to fetch signed urls when the document is non-public, on the details pages.


api/src/utils/put-utils.js
```diff
@@ -112,3 +114,71 @@ exports.getDotNotation = function(obj, target, prefix) {
...
+  if (!sanitizedObj) {
+    return updateObj;
+  }
+
+  // apply anonymous business logic
+  if (QueryUtils.isRecordConsideredAnonymous(sanitizedObj)) {
+    updateObj.$pull['issuedTo.read'] = 'public';
+  } else {
+    updateObj.$addToSet['issuedTo.read'] = 'public';
+  }
```
marklise, line 181: Maybe passing in the object property instead of hardcoding here would be better:

`updateObj.$pull['issuedTo.read'] = 'public'; ` => `updateObj.$pull[someProperty] = 'public';`

I don't know if QueryUtils is where the real business logic should sit.  I personally like to see the consideration for the reading/modification of the roles, and determining anonymous/not anonymous in the same file.  Not spread out in put-utils, and not in query-utils, but in something like businessManager.js.  This way all the business logic is in a single location to be reviewed independently of other standard functions, like queryUtils, putUtils, and PostUtils.

api/src/integrations/nris/datasource.js
```diff
@@ -195,7 +185,7 @@ class NrisDataSource {
     newRecord.addedBy = (this.auth_payload && this.auth_payload.displayName) || '';
     newRecord.updatedBy = (this.auth_payload && this.auth_payload.displayName) || '';
 
-    newRecord.sourceSystemRef = 'nris';
+    newRecord.sourceSystemRef = 'nris-epd';
```
marklise, line 188: We should be updating the swagger dataSourceType enum, as well as the task service type (currently 'nris'), and the getDataSourceConfig switch handler.

NickPhura: Yeah good call.  The swagger is a bit behind, and only mentions orders and inspections, so I can update that as well.

api/src/utils/put-utils.js
NickPhura: Yeah I can move all of the business logic to a single file, thats a good idea. 

Not sure about passing the property in.  The `applyBusinessLogic`  functions are currently black-boxes that apply any/all business logic to the object/record.  Currently we only have the anonymous logic, but in theory it could be other stuff that is custom to the business.  I could totally see a future enhancement to this, where a config is passed in that tells it which rules to run or skip, and maybe what fields to apply each rule to, etc.  But right now the logic is all the same for each record type, and they all must apply it.

NickPhura:
> 
> 
> Noticed a bug with issued to. Not sure if its related to this PR though. You can see it here: https://bcmines.atlassian.net/browse/PTI-590

PR to fix this: https://github.com/bcgov/NRPTI/pull/352

api/src/utils/put-utils.js
marklise: Looks good.  Most important I think was to move it into a single location.

NickPhura:
@marklise Regarding your last comment:

I believe that particular record falls under the nris  exception, which prevents it from ever being anonymous.

[business-logic-manager.js](https://github.com/bcgov/NRPTI/blob/472bf62b426f07a7febc9670077a4fa8a70ab720/api/src/utils/business-logic-manager.js) Lines 65-73

marklise:
We should probably add some UI/UX in the future around messaging the administrator that these sorts of publishing rules aren't applied to these records.

NickPhura:
> 
> 
> We should probably add some UI/UX in the future around messaging the administrator that these sorts of publishing rules aren't applied to these records.

For sure.  We do have a ticket in the backlog to add some messaging to the admin UI.  Currently it w as just to tell the user that its anonymous, etc.  But could be expanded to tell them why its anonymous too.

PR #360, opened 2020-05-12
PTI-601: Tweak anonymous business logic

NickPhura:
- During a put, if the issuedTo object is not present because it was not changed, don't update anonymity from whatever it was before.


api/src/utils/business-logic-manager.js
```diff
@@ -36,7 +36,8 @@ exports.applyBusinessLogicOnPost = function(record) {
   }
 
   // apply anonymous business logic
-  if (isRecordConsideredAnonymous(record)) {
+  const isAnonymous = isRecordConsideredAnonymous(record);
+  if (isAnonymous === null || isAnonymous === true) {
```
marklise, line 40: I think the check for issuedTo should be outside this function, not inside it.

NickPhura: Do you mean the check for whether or not the issuedTo object is even present?
- We couldn't pull it out given the current way the business logic rules are applied.  Currently, this `applyBusinessLogicOnPost` function runs all of the business logic rules and applies all of the necessary changes.  If we wanted to pull part of the issuedTo check out, we would need to scrap this function, and instead run all of the rules from the controllers.  And then each controller would be responsible for running all of the appropriate rules, and applying the appropriate changes, themselves.

marklise: I was thinking more like this:

`function isIssuedToConsideredAnonymous(record)`
=>
`function isIssuedToConsideredAnonymous(issuedTo)`

and

`const isAnonymous = isRecordConsideredAnonymous(record);`
=>
`const isAnonymous = record.issuedTo !== null/undefined ? isRecordConsideredAnonymous(record.issuedTo) : true/false`

Also, I think eventually we *will* need controllers specifying exactly what is anonymous and what isn't depending on the type of record coming in.  We'll probably run into this problem sooner or later.

NickPhura: > I was thinking more like this:
> 
> `function isIssuedToConsideredAnonymous(record)`
> =>
> `function isIssuedToConsideredAnonymous(issuedTo)`
> 
> and
> 
> `const isAnonymous = isRecordConsideredAnonymous(record);`
> =>
> `const isAnonymous = record.issuedTo !== null/undefined ? isRecordConsideredAnonymous(record.issuedTo) : true/false`

Gotcha, yeah that makes sense. Ill make that change.


> Also, I think eventually we _will_ need controllers specifying exactly what is anonymous and what isn't depending on the type of record coming in. We'll probably run into this problem sooner or later.

Yeah absolutely, as soon as the record types diverge, having some of this logic all in one place will start to make less sense.

api/src/utils/business-logic-manager.js
```diff
@@ -56,7 +59,11 @@ exports.applyBusinessLogicOnPost = function(record) {
  * @returns boolean true if the record is considered anonymous, false otherwise.
  */
 function isRecordConsideredAnonymous(record) {
-  let isAnonymous = isIssuedToConsideredAnonymous(record);
+  if (!record) {
+    return null;
+  }
```
marklise, line 64: Boolean functions shouldn't be returning null.

PR #368, opened 2020-05-13
PTI-583: Update ocers-data.csv

NickPhura:
- Also update the subSection and paragraph regex to be more robust.

### DIFF
--- a/api/migrations/20200211065122-nrcedData.js
+++ b/api/migrations/20200211065122-nrcedData.js
@@ -127,8 +127,8 @@ const createAdministrativePenalty = async function (row, nrptiCollection) {
       act: row[7],
       regulation: row[8],
       section: row[9],
-      subSection: row[10].replace(/[()]/g, ''),
-      paragraph: row[11].replace(/[()]/g, '').replace(/[()]/g, '')
+      subSection: cleanSubSectionOrParagraph(row[10]),
+      paragraph: cleanSubSectionOrParagraph(row[11])
     },
     issuedTo: getIssuedToObject(row, true),
     // projectName: '',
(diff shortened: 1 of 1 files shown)


api/migrations/20200211065122-nrcedData.js
```diff
@@ -972,6 +972,21 @@ const calculateDateFormat = function (row) {
...
+  // remove a `-` or `(` at the start, and a `)` at the end, if present.
+
+  // Why remove brackets? SubSections and Paragraphs are visually displayed inside brackets (ie: (a) ) but are currently
+  // stored in mongo without them. The brackets are then added back in the frontend UI.
+
+  // Why remove leading dashes?  Excel likes to display `(1)` but then secretly export `-1`.  Admin users are unlikely
+  // to be aware of this, so handling it here is safest.
+  return value.replace(/^[-(]/, '').replace(/[)]$/,'');
+}
+
```
marklise, line 989: Shouldn't we be replacing all occurrences of non-numerics?

NickPhura: Yes and no.

The most common subSections/Paragraphs are of the form:
`a`, `i`, `1`
but there are also instances of:
`1.2`, `a-c`

So we would need to be a bit careful about stripping out all dashes, for example.

I suspect this type of issue will be re-occurring, as more types of CSVs are introduced, and which might have other strange values.

I've yet to see an instance of `(a)(b)` (and maybe there won't ever be), but in this kind of case it would be more awkward to figure out how to handle things properly.

NickPhura: In other words, not entirely sure what the best solution is here

marklise: Right, I should have said non-alpha, not non-numerics.  I have yet to see 'a-c', but certainly 1.2 or f.3 exists.  I would think a global removal of -, (, ), would be appropriate here?  Do you have an example of a-c?  Which act has that?

PR #369, opened 2020-05-13
PTI-607: unhide agency filter

NickPhura:
Apparently the mockup that showed that the agency filter should be hidden was out of date, so unhiding it.


NickPhura:
Closing as very few records have issuingAgency, and therefore it is not an accurate or useful filter for the public.

PR #375, opened 2020-05-20
NRPT-64: Add new migration for additional ocers csv data.

NickPhura:
#### Commit 1
Add additional ocers csv and migration script.

The migration script is exactly the same as the original ocers (nrced) migration, with only the csv filename being different.

#### Commit 2
Noticed a discrepancy in the date format check.  I didn't update the old ocers migration, as I'm not sure if that is okay or not, but also because I did some checking and that date format isn't found in the old csv anyways.  Let me know what you think though.


api/migrations/20200519165645-ocersData.js
```diff
@@ -0,0 +1,996 @@
...
+  penaltyObject.description = row[15] || '';
+
+  return [penaltyObject];
+};
+
+const calculateDateFormat = function (row) {
+  let dateIssued = null;
+  if (row[1]) {
+    if (moment(row[1], 'MM/DD/YYYY', true).isValid()) {
+      dateIssued = moment(row[1], 'DD/MM/YYYY').toDate();
```
marklise: Why are we checking MM/DD/YYYY but then creating DD/MM/YYYY?

NickPhura: This is the date format issue I commented about in the description of the PR. 

It was an issue in the original nrced csv migration, which I fixed in the new ocers migration, in the second commit.

PR #378, opened 2020-05-28
NRPT-70: Admin Projects Filter

NickPhura:
### Commit 1
- Add Projects advanced filters,

### Commit 2
- Add unit tests.
  - I put a bit of a timebox on the tests, since many of these classes don't have any at the moment, and I could easily spend over a week just writing tests.  I focused on testing only the code that was new or directly touched as a result of this feature, in the 2 spots that are the most impactful and most likely to run into issues if changes are made in future.

### Commit 3
- Linted the files that were touched.

### Commit 4
- Code review updates

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
+++ b/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
@@ -143,7 +143,8 @@ export class RecordsListComponent implements OnInit, OnDestroy {
           this.queryParams['act'] ||
           this.queryParams['regulation'] ||
           this.queryParams['sourceSystemRef'] ||
-          this.queryParams['hasDocuments']
+          this.queryParams['hasDocuments'] ||
+          this.queryParams['projects']
         ) {
           this.showAdvancedFilters = true;
         }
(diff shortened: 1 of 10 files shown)


api/src/controllers/search.js
```diff
@@ -15,86 +15,142 @@ function isEmpty(obj) {
...
-      })
-    );
+/**
+ * Generate an array of expressions for the query string parameters.
+ *
+ * @param {string} field query string
+ * @param {string} [logicalOperator='$or'] mongo logical operator ('$or', '$and')
+ * @param {string} [comparisonOperator='$eq'] mongo comparison operator ('$eq', '$ne')
+ * @returns {object[]} array of objects
+ */
```
mtCarto, line 25: Glad to see JS Docs 👍

angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
```diff
@@ -433,6 +454,29 @@ export class RecordsListComponent implements OnInit, OnDestroy {
...
+   * @memberof RecordsListComponent
+   */
+  getProjectsFilterArray(projects: object): string[] {
+    if (!projects) {
+      return [];
+    }
+
+    const projectsQueryParam: string[] = [];
+
+    for (const projectName of Object.keys(projects)) {
```
KitArmstrong, line 471: You can use `for...in` to enumerate over an object's properties instead of 'for...of` and then you don't need to use `Object.keys()`.

angular/projects/admin-nrpti/src/app/records/records-resolver.ts
```diff
@@ -43,43 +43,73 @@ export class RecordsResolver implements Resolve<Observable<object>> {
       subset = params.subset.split(',');
     }
 
-
-    const filterParams = {};
+    const and = {};
```
KitArmstrong, line 46: This is much clearer.

angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
NickPhura: I think I've read somewhere that for...in is not preferred, as it has to make a bunch of extra calls and will grab inherited values and doesn't provide function scope.  (Obviously this use case is super simple, so it probably doesn't matter either way).  I'd probably be inclined to leave it as is, unless my memory is totally wrong

KitArmstrong: Totally fine, both work well.

angular/projects/admin-nrpti/src/app/records/records-resolver.ts
```diff
@@ -43,43 +43,73 @@ export class RecordsResolver implements Resolve<Observable<object>> {
...
+            // Other + LNG Canada is equivalent to NOT Coastal Gaslink
+            nor['projectName'] = 'Coastal Gaslink';
+          } else {
+            // Other + Coastal Gaslink is equivalent to NOT LNG Canada
+            nor['projectName'] = 'LNG Canada';
+          }
+        }
+      } else if (projectNames.length) {
+        or['projectName'] = projectNames.join(',');
+      }
```
marklise, line 112: This is a string lookup, which is not indexed and will be slow as a result.  Instead, search on _epicProjectId: ObjectId("588511c4aaecd9001b825604") or ObjectId("5885101caaecd9001b80b41f").

NickPhura: Updated!

PR #381, opened 2020-05-29
NRPT-69: Updated post to add 2 way reference between master and flavours

danieltruong:


api/src/controllers/post/administrative-penalty.js
```diff
@@ -32,53 +33,67 @@ const BusinessLogicManager = require('../../utils/business-logic-manager');
...
+    const db = mongodb.connection.db(process.env.MONGODB_DATABASE || 'nrpti-dev');
+    const collection = db.collection('nrpti');
+    let orArray = [];
+    for (let i = 0; i < idsToDelete.length; i++) {
+      orArray.push({ _id: new ObjectId(idsToDelete[i]) });
+    }
+    await collection.deleteMany({
+      $or: orArray
+    });
+
```
marklise, line 86: I think we can reduce this to a common function in order to reduce the copy-paste going on here.

danieltruong: ec1f4f4d0936283400ecee9070a336d5dafecd53

PR #384, opened 2020-06-02
NRPT-62: Remove ocers-data csv files (move to confluence)

NickPhura:
Fixed a misalignment between ocers csv versions in the migrations.
- This ocers csv in this PR aligns with what is in Prod currently (as it should), and aligns correctly with the additional records imported by the second newer ocers migration.

See Kyles comment about what happened here: https://bcmines.atlassian.net/browse/NRPT-64


NickPhura:
@marklise From our conversation earlier (let me know if I missed anything important):

Locally

1. I dumped the prod DB
2. I dropped my local
3. Restored prod into my local
4. Ran the migrations as they exist currently in Test.
5. Verified that the number of records aligns with our expectations, and that some of the duplicates that had been found in Test by Kyle only existed once now

1. Followed the same steps 1-3 as above.
2. Ran the migrations as they exist including this PR
3. Verified that the number of records aligns with our expectations, and that some of the duplicates that had been found in Test by Kyle only existed once now

Dev

1. I dropped the Dev database
2. Restored Prod into Dev.
3. I ran the migrations against Dev.
4. I asked Kyle to review the data.

NickPhura:
@marklise 

Regarding my previous comment: Kyle took a look at the data, and verified that it is correct.

He did find 5 records in the ocers-data-2.csv that had minor mistakes (names that should have been redacted). So I've added the new version of ocers-data-2.csv to this PR.  The new ocers-data-2.csv never made it past Test, so it should be safe to update.

NickPhura:
As per the discussions around storing csv data: I've updated this PR to remove the remaining ocers-data csv.  

I've created a page in confluence and uploaded both ocers-data csvs: https://apps.nrs.gov.bc.ca/int/confluence/display/MDS/NRPTI+Database+Migrations

PR #390, opened 2020-06-03
NRPT-122: Initial mine list page + related files.

NickPhura:
Initial mines list page + related module/routing files. 

I've included only the bare minimum set of tests, as there is not much functionality to test yet, and seemed more prudent to get these initial components merged so they aren't blocking anyone else.

I did do a bit of manual testing, by manually creating a mine record, and checking that the table loaded correctly.


NickPhura:
> 
> 
> Does as advertised. Nothing stands out and seems to match up with styles of other tables. Paging/sorting etc. all work as expected.
> 
> Is there any need/plans to include a keyword search for this screen as well?

I'm guessing there will be some amount of search/advanced search added eventually.  I just didnt bother porting it from the existing records list page since we don't know what the mine filters will look like (what fields would need text indexes, etc).

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.ts
```diff
@@ -0,0 +1,210 @@
...
+            res.mines[0].data.meta[0].searchResultsTotal) ||
+          0;
+
+        this.tableData.columns = this.tableColumns;
+
+        this._changeDetectionRef.detectChanges();
+        this.loadingScreenService.setLoadingState(false, 'body');
+      });
+    });
+  }
```
marklise, line 122: You have two subscriptions with data being set, but only one detectChanges.  Also redirecting to / is a bit harsh, showing 'no records found' along with the error in the page view instead of flicking around would probably be better UX.

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  public tableData: TableObject = new TableObject({ component: MinesTableRowComponent });
+  public tableColumns: IColumnObject[] = [
+    {
+      name: 'Name',
+      value: 'name',
+      width: 'col-3'
+    },
+    {
+      name: 'owner',
```
marklise, line 41: Capitalize

dhlevi:
There's some model changes indicated in the [PR 393](https://github.com/bcgov/NRPTI/pull/393). Specifically: https://github.com/bcgov/NRPTI/pull/393/files#diff-d33c850871eaa41feb49cec3d912d8d5

Not sure about the schema naming, but if we're changing `owner` to `permitee` should probably change the header/ref in the table when they merge in.

NickPhura:
> 
> 
> There's some model changes indicated in the [PR 393](https://github.com/bcgov/NRPTI/pull/393). Specifically: https://github.com/bcgov/NRPTI/pull/393/files#diff-d33c850871eaa41feb49cec3d912d8d5
> 
> Not sure about the schema naming, but if we're changing `owner` to `permitee` should probably change the header/ref in the table when they merge in.

Yeah for sure.  This is where order of merges makes it a bit tricky though. I could per-emptively update this in my PR, but then technically it would be broken, if only temporarily. But also, we won't have any data to load either, so not sure :/

marklise:
> > There's some model changes indicated in the [PR 393](https://github.com/bcgov/NRPTI/pull/393). Specifically: https://github.com/bcgov/NRPTI/pull/393/files#diff-d33c850871eaa41feb49cec3d912d8d5
> > Not sure about the schema naming, but if we're changing `owner` to `permitee` should probably change the header/ref in the table when they merge in.
> 
> Yeah for sure. This is where order of merges makes it a bit tricky though. I could per-emptively update this in my PR, but then technically it would be broken, if only temporarily. But also, we won't have any data to load either, so not sure :/

We don't want to be messing with 'orders' of merges.  If a PR can't go in on it's own, then it shouldn't land period, and it should not be submitted for review.  There's nothing wrong with an iterative state of the code as long as it doesn't break anything or cause other issues down the road.

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.ts
NickPhura: Capitalize: will fix.

The sorting by name is working (though I will change it to `+name` instead of `-name`), it is just the UI column arrow that isn't working.

The parameters in the url don't show up initially, but do appear once you click any of the pagination controls.

It looks like the problem is that any parameters the resolver uses, but which aren't in the url already (IE: default params), don't get propagated to the list component.  So the data resolves with filters/sorting applied as we expect, but the component has no knowledge of it, and so it loads with no params in the url, and no UI indications (like sorting arrows).

So these issues might warrant their own ticket, as we probably want to figure out the best way to do this, and then update all of our list components/resolvers.

> 
> 
> > > There's some model changes indicated in the [PR 393](https://github.com/bcgov/NRPTI/pull/393). Specifically: https://github.com/bcgov/NRPTI/pull/393/files#diff-d33c850871eaa41feb49cec3d912d8d5
> > > Not sure about the schema naming, but if we're changing `owner` to `permitee` should probably change the header/ref in the table when they merge in.
> > 
> > 
> > Yeah for sure. This is where order of merges makes it a bit tricky though. I could per-emptively update this in my PR, but then technically it would be broken, if only temporarily. But also, we won't have any data to load either, so not sure :/
> 
> We don't want to be messing with 'orders' of merges. If a PR can't go in on it's own, then it shouldn't land period, and it should not be submitted for review. There's nothing wrong with an iterative state of the code as long as it doesn't break anything or cause other issues down the road.

Yeah exactly.  I was going to leave it, and then just sync up with Kit about one of us updating it, depending on whos stuff goes in first.

NickPhura:
> 
> 
> > > There's some model changes indicated in the [PR 393](https://github.com/bcgov/NRPTI/pull/393). Specifically: https://github.com/bcgov/NRPTI/pull/393/files#diff-d33c850871eaa41feb49cec3d912d8d5
> > > Not sure about the schema naming, but if we're changing `owner` to `permitee` should probably change the header/ref in the table when they merge in.
> > 
> > 
> > Yeah for sure. This is where order of merges makes it a bit tricky though. I could per-emptively update this in my PR, but then technically it would be broken, if only temporarily. But also, we won't have any data to load either, so not sure :/
> 
> We don't want to be messing with 'orders' of merges. If a PR can't go in on it's own, then it shouldn't land period, and it should not be submitted for review. There's nothing wrong with an iterative state of the code as long as it doesn't break anything or cause other issues down the road.

Yeah exactly.  I was going to leave it, and then just sync up with Kit about one of us updating it, depending on whos stuff goes in first.

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.ts
NickPhura: > Capitalize
Updated

> Mines list page isn't sorting by name by default, and isn't setting the parameters for the URL properly. All the various pagination and table sorting should be showing up in the URL.
Updated to account for initial/default url parameters.

NickPhura:
> 
> 
> The url is using the wrong type of parameters, we are using matrix params.
> 
> PR:
> image
> 
> List page:
> 
> image

Updated

angular/projects/admin-nrpti/src/app/mines/mines-rows/mines-table-row.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    switch (this.rowData._schemaName) {
+      case 'Mine':
+        this.router.navigate(['mines', 'mines', this.rowData._id, 'edit']);
+        break;
+      default:
+        break;
+    }
+  }
+}
```
marklise, line 53: Not sure why we're switching here when an if statement will do.

NickPhura: This is the pattern we used in other places, which I copied over.
Can change it if its important

marklise: I don't think switches with a single item make sense.  It's not really achieving anything other than adding to vertical space wasted.

NickPhura: sure, will put up a fix in 2 mins

NickPhura: updated

NickPhura:
> 
> 
> One last nit - I notice the hover state (showPointer class) is activated only on each TD, instead of the TR.
> 
> This can land after that nit is fixed.

Updated.

This also exists in our records list and news list pages, so I can make a bug ticket to address them as well.

PR #392, opened 2020-06-03
BCMI data update migration

dhlevi:
Migration script to fetch data from the BCMI api and update matching records in NRPTI. 

We don't currently have any data from Core, so testing was done by creating some mock data with matching names (basically just copied some data in from BCMI directly and matched the schema)

Schema name is `MineBCMI` in the script, matching with [PR 338](https://github.com/bcgov/NRPTI/pull/388). This can be updated as needed! sourceSystemRef for all updated records will be set to `mem-admin` so it should be easy to wipe and rerun the migration as needed.

See ticket [NRPT-102](https://bcmines.atlassian.net/browse/NRPT-102)


api/migrations/20200603153526-bcmiDataLoad.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    return nrpti.update({ _id: new ObjectID(nrptiMines[0]._id) }, {
+      $set: {
+        type:            mineData.type,
+        summary:         mineData.description, // BCMI doesn't have a "summary" attribute
+        description:     mineData.description,
+        links:           externalLinks,
+        updatedBy:       'NRPTI BCMI Data Migration',
+        sourceSystemRef: 'mem-admin'
```
NickPhura, line 101: I don't think we want to change the sourceSystemRef here.  

In this case it will have been set to 'core' by the core importer.  But for any of these records, it should reflect whichever 'source' system it was imported from, and which it will receive periodic updates from.

marklise: I disagree - mem-admin is the sourceSystemRef since this is where the data is coming from.  It's not coming from core, but from mem-admin service: https://mines.empr.gov.bc.ca/

NickPhura: Maybe I am misunderstanding the purpose of this migration.  The code indicates that it is only updating an existing mine record.  Where does the original mine record come from? I assumed it was Core, based on the ticket description, but maybe that isn't the case.

marklise: It's just what data source endpoint this came from.  We say epic because we're calling epic api and that's where we pulled that data from, same for nris-epd.  We also store refId's based on the called system, as we don't really know if that system created it or pulled it in from somewhere else.

NickPhura: I guess this is mainly what I'm wondering:

If we run the core importer and create a new record from core, and then update it via migration from bcmi, and then update it again in future by running the core importer again, would the sourceSystemRef change each time (core -> mem-admin -> core) or would it stay fixed as core?

PR #393, opened 2020-06-04
Adds Core API integration

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-109

Adds an importer that retrieves new and updated records from Core. The importer was created to be run as a system job rather than one that is triggered by the user. The CRON is not part of this ticket as there is another ticket to handle creating the scheduling.

Note:

There are no tests. Will be adding those in a separate PR, but wanted to get this in to unblock others
There will most likely need to be some changes to a mine's commodities, but it is currently fetching that data the best way we know how
The last AC on the ticket mentions providing error feedback to the user. This is done through logging as this is not a user-triggered job

### DIFF
--- a/api/src/controllers/post/mine.js
+++ b/api/src/controllers/post/mine.js
@@ -1,0 +1,68 @@
+const mongoose = require('mongoose');
+
+const postUtils = require('../../utils/post-utils');
+
+/**
+ * Create a new Mine record.
+ *
+ * @param {*} args
+ * @param {*} res
+ * @param {*} next
+ * @param {*} incomingObj Mine record to create
+ * @returns new Mine record
+ */
+exports.createRecord = async function (args, res, next, incomingObj) {
+  // TODO: Populate if needed.
+  const flavourFunctions = {};
+  return await postUtils.createRecordWithFlavours(args, res, next, incomingObj, this.createMaster, flavourFunctions);
+};
+
+/**
+ * Performs all operations necessary to create a master Mine record.
+ *
+ * @param {*} args
+ * @param {*} res
+ * @param {*} next
+ * @param {*} incomingObj Mine record to create
+ * @param {*} flavourIds array of flavour record _ids
+ * @returns created master Mine record
+ */
+exports.createMaster = function (args, res, next, incomingObj, flavourIds) {
+  let Mine = mongoose.model('MineBCMI');
+  let mine = new Mine();
+
+  mine._schemaName = 'MineBCMI';
+
+  // set permissions
+  mine.read = ['sysadmin'];
+  mine.write = ['sysadmin'];
+
+  // set data
+  incomingObj.name && (mine.name = incomingObj.name);
+  incomingObj.permitNumbers.length && (mine.permitNumbers = incomingObj.permitNumbers);
+  incomingObj.status && (mine.status = incomingObj.status);
+  incomingObj.commodities.length && (mine.commodities = incomingObj.commodities);
+  incomingObj.region && (mine.region = incomingObj.region);
+  incomingObj.location && (mine.location = incomingObj.location);
+  incomingObj.permittee && (mine.permittee = incomingObj.permittee);
+  incomingObj.type && (mine.type = incomingObj.type);
+  incomingObj.summary && (mine.summary = incomingObj.summary);
+  incomingObj.description && (mine.description = incomingObj.description);
+  incomingObj.links.length && (mine.links = incomingObj.links);
+
+  // Not checking value as it could be 0 which would fail the falsey check.
+  mine.tailingImpoundments = incomingObj.tailingImpoundments;
+
+  // Set meta. If the args exist then use the auth otherwise check the incoming object. This occurs if 
+  // the system is creating the record.
+  mine.addedBy = args && args.swagger.params.auth_payload.displayName || incomingObj.addedBy;
+  mine.updatedBy = args && args.swagger.params.auth_payload.displayName || incomingObj.updatedBy;
+  mine.dateAdded = new Date();
+  mine.dateUpdated = new Date();
+
+  // set data source reference
+  incomingObj.sourceSystemRef && (mine.sourceSystemRef = incomingObj.sourceSystemRef);
+  incomingObj._sourceRefId && (mine._sourceRefId = incomingObj._sourceRefId);
+
+  return mine;
+};
--- a/api/src/controllers/post/mine.js
+++ b/api/src/controllers/post/mine.js
@@ -28,10 +28,10 @@ exports.createRecord = async function (args, res, next, incomingObj) {
  * @returns created master Mine record
  */
 exports.createMaster = function (args, res, next, incomingObj, flavourIds) {
-  let Mine = mongoose.model('MineBCMI');
+  let Mine = mongoose.model('Mine');
   let mine = new Mine();
 
-  mine._schemaName = 'MineBCMI';
+  mine._schemaName = 'Mine';
 
   // set permissions
   mine.read = ['sysadmin'];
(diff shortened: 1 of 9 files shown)


api/src/controllers/post/mine.js
```diff
@@ -0,0 +1,68 @@
...
+  // Set meta. If the args exist then use the auth otherwise check the incoming object. This occurs if 
+  // the system is creating the record.
+  mine.addedBy = args && args.swagger.params.auth_payload.displayName || incomingObj.addedBy;
+  mine.updatedBy = args && args.swagger.params.auth_payload.displayName || incomingObj.updatedBy;
+  mine.dateAdded = new Date();
+  mine.dateUpdated = new Date();
+
+  // set data source reference
+  incomingObj.sourceSystemRef && (mine.sourceSystemRef = incomingObj.sourceSystemRef);
+  incomingObj._sourceRefId && (mine._sourceRefId = incomingObj._sourceRefId);
```
NickPhura, line 65: Our other post controllers have some extra handling around the objectIds that we should continue here as well.  I can't remember exactly, but I think its to avoid any chance that mongo doesn't recognize them as ObjectIds, which messes with search queries, etc.
```
  incomingObj._sourceRefId &&
    ObjectId.isValid(incomingObj._sourceRefId) &&
    (inspection._sourceRefId = new ObjectId(incomingObj._sourceRefId));
```

api/src/controllers/post/mine.js
```diff
@@ -0,0 +1,68 @@
...
+  incomingObj.permitNumbers.length && (mine.permitNumbers = incomingObj.permitNumbers);
+  incomingObj.status && (mine.status = incomingObj.status);
+  incomingObj.commodities.length && (mine.commodities = incomingObj.commodities);
+  incomingObj.region && (mine.region = incomingObj.region);
+  incomingObj.location && (mine.location = incomingObj.location);
+  incomingObj.permittee && (mine.permittee = incomingObj.permittee);
+  incomingObj.type && (mine.type = incomingObj.type);
+  incomingObj.summary && (mine.summary = incomingObj.summary);
+  incomingObj.description && (mine.description = incomingObj.description);
+  incomingObj.links.length && (mine.links = incomingObj.links);
```
NickPhura, line 51: We probably have to check that links is not null before calling .length, otherwise it could throw an error.
Same with the other .length calls.

```incomingObj.links && incomingObj.links.length ...```

File: api/src/controllers/post/mine.js [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+ * @param {*} next
+ * @param {*} incomingObj Mine record to create
+ * @param {*} flavourIds array of flavour record _ids
+ * @returns created master Mine record
+ */
+exports.createMaster = function (args, res, next, incomingObj, flavourIds) {
+  let Mine = mongoose.model('MineBCMI');
+  let mine = new Mine();
+
+  mine._schemaName = 'MineBCMI';
```
NickPhura, line 34: For master records, the model/schema is `Mine`.

The flavour record will be `MineBCMI`

api/src/controllers/put/mine.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Performs all operations necessary to edit a master Mine record.
+ *
+ * @param {*} args
+ * @param {*} res
+ * @param {*} next
+ * @param {*} incomingObj see example
+ * @returns object containing the operation's status and created records
+ */
+exports.editRecord = async function(args, res, next, incomingObj) {
```
NickPhura, line 15: In keeping with the other controllers, this should be called `editMaster` and then there should be a separate `editRecord` function that makes all the calls needed to do all the editing. I think Daniel has an open PR where he changed some of this though, so depending on merge order, might be a bit weird to sync up on.

api/src/controllers/put/mine.js
```diff
@@ -0,0 +1,57 @@
...
+  try {
+    if (incomingObj._schemaName !== 'MineBCMI') {
+      throw new Error('editRecord - incorrect schema type, must be MineBCMI');
+    }
+
+    const Model = mongoose.model('MineBCMI');
```
NickPhura, line 21: Similar comment to the mine Post constoller: 

For master records, the model/schema is Mine.

The flavour record will be MineBCMI

api/src/controllers/put/mine.js
```diff
@@ -0,0 +1,57 @@
...
+
+    const record = await Model.findOneAndUpdate(
+        { _id: new ObjectId(incomingObj._id) },
+        { $set: {
+            name: incomingObj.name,
+            permitNumbers: incomingObj.permitNumbers,
+            status: incomingObj.status,
+            commodities: incomingObj.commodities,
```
NickPhura, line 29: Rather than manually setting every value, there is some santize/dotnotation code (see other put controllers) that make it a bit simpler to safely set everything, including nested objects.

api/src/controllers/post/mine.js
KitArmstrong: I did see that. The reason it isn't used here is the source system is Postgres, not Mongo. The ID field is a UUID, which is not a format that works with Mongo's Object ID. I think it will have to be saved as a string unless someone knows a better way.

api/src/controllers/post/mine.js
KitArmstrong: Updated

api/src/controllers/post/mine.js
KitArmstrong: Updated. I wasn't 100% sure about the whole flavour thing when writing this.

api/src/controllers/put/mine.js
KitArmstrong: Updated to use existing helper functions.

api/src/controllers/put/mine.js
KitArmstrong: Updated

api/src/controllers/put/mine.js
KitArmstrong: Updated to match the structure of the other controllers.

api/src/controllers/post/mine.js
NickPhura: Ah gotcha, that makes sense.

api/src/controllers/put/mine.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {*} args
+ * @param {*} res
+ * @param {*} next
+ * @param {*} incomingObj
+ * @returns object containing the operation's status and created records
+ */
+exports.editRecord = async function(args, res, next, incomingObj) {
+  try{
+    // TODO: Add any flavour actions here.
+    console.log('here');
```
NickPhura, line 18: nit: console log

KitArmstrong: Woops. Gone!

api/src/controllers/post/mine.js
marklise: _sourceRefId is just a convenience property meant to keep track of the unique identifier in the source system.  It shouldn't have to be an ObjectId as mentioned by @KitArmstrong.  There is no requirement in MongoDB that all '_sourceRefId' instances be the same datatype for every single database document (This is counter to how SQL-based databases work).  The code, in it's safest/robust form would inspect the sourceSystemRef, and determine whether to wrap in an ObjectId, or treat as a string.

This is one of the issues with using Mongoose and having our schema slightly different in some instances of a schema.  It may be better to convert all _sourceRefId to string.

api/src/integrations/core/datasource.js
```diff
@@ -0,0 +1,320 @@
+'use strict';
+
+const axios = require('axios');
+const QS = require('qs');
+
+const integrationUtils = require('../integration-utils');
+const MineUtils = require('./mines-utils');
```
marklise, line 7: mines-utils => mine-utils

KitArmstrong:
300 was about all the Test API would handle. Anything closer to 500 and it would time out. So I chose a number I thought would be fairly safe, although might need some tweaking.

dhlevi:
> 300 was about all the Test API would handle. Anything closer to 500 and it would time out. So I chose a number I thought would be fairly safe, although might need some tweaking.

Yeah makes sense. Might be reasonable to use an environment var if it's going to need some flexibility so we're not necessarily releasing code if it needs to change quickly. Too small a number though and it'll probably look like we're trying to DDoS their api!

KitArmstrong:
> Yeah makes sense. Might be reasonable to use an environment var if it's going to need some flexibility so we're not necessarily releasing code if it needs to change quickly. Too small a number though and it'll probably look like we're trying to DDoS their api!

Good point. I updated to use an environment variable or default to 300 if one is not set.
>

PR #394, opened 2020-06-04
NRPT-146: Record edit cleanup and optimization

danieltruong:


api/src/utils/put-utils.js
```diff
@@ -112,3 +114,122 @@ exports.getDotNotation = function(obj, target, prefix) {
...
+  delete flavourIncomingObj._flavourRecords;
+  delete flavourIncomingObj._master;
+  delete flavourIncomingObj.read;
+  delete flavourIncomingObj.write;
+
+  // Prepare flavours
+  const entries = Object.entries(flavourFunctions);
+  // Example of entries: [['OrderLNG', createLNG()], ['OrderNRCED', createNRCED()]]
+
+  if (entries.length > 0 && Object.keys(flavourIncomingObj).length > 0) {
```
NickPhura, line 137: This function, and the similar one in the post utils, are prime candidates from some cost effective unit tests.  With all of the nested conditionals and loops, they would be difficult for a new dev to debug/update if they had to.

PR #397, opened 2020-06-05
NRPT-102: BCMI data update script

dhlevi:
Migration script to fetch data from the BCMI api and update matching records in NRPTI. 

We don't currently have any data from Core, so testing was done by creating some mock data with matching names (basically just copied some data in from BCMI directly and matched the schema)

Schema name is `MineBCMI` in the script, matching with [PR 338](https://github.com/bcgov/NRPTI/pull/388). This can be updated as needed! sourceSystemRef for all updated records will be set to `mem-admin` so it should be easy to wipe and rerun the migration as needed.

See ticket [NRPT-102](https://bcmines.atlassian.net/browse/NRPT-102)

### DIFF
--- a/api/migrations/20200603153526-bcmiDataLoad.js
+++ b/api/migrations/20200603153526-bcmiDataLoad.js
@@ -1,0 +1,146 @@
+'use strict';
+
+let dbm;
+let type;
+let seed;
+
+const https = require('https');
+// register the mine schema/model
+const ObjectID = require('mongodb').ObjectID;
+const bcmiUrl = 'https://mines.empr.gov.bc.ca'; // prod api url
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  let mClient;
+
+  try {
+    mClient = await db.connection.connect(db.connectionString, { native_parser: true });
+    const nrpti = mClient.collection('nrpti');
+
+    // API call to pull data from BCMI
+    // could also fetch from NRPTI first: require('../src/models/bcmi/mine').find().then(...);
+    // then match to BCMI, just doing it this way so we can match on name, rather then code
+    console.log('Fetching all major mines in BCMI...');
+    // 30 results for major, 74 for published. Note, published does not return links so if we're supposed to use
+    // published, we'll need to do a follow up call to /api/project/bycode/<mineData.code> to get the links
+    const publishedMines = await getRequest(bcmiUrl + '/api/projects/published');
+    console.log('Located ' + publishedMines.length + ' mines. Batching updates...');
+
+    const promises = [];
+    // build up a collection of all requests
+    for (let i = 0; i < publishedMines.length; i++) {
+      try {
+        // The published endpoint doesn't have links, refresh the mineData object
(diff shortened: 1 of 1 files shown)


api/migrations/20200603153526-bcmiDataLoad.js
```diff
@@ -0,0 +1,146 @@
...
+    // Alternatively...
+    /* let nrptiMine = await require('../src/models/bcmi/mine').findById(nrptiMines[0]._id);
+
+    nrptiMine.type        = mineData.type;
+    nrptiMine.summary     = mineData.description;
+    nrptiMine.description = mineData.description;
+    nrptiMine.links       = externalLinks;
+    nrptiMine.updatedBy   = 'NRPTI BCMI Data Migration';
+
+    return nrptiMine.save(); */
```
NickPhura, line 114: May as well remove, if we aren't going this route.

dhlevi: Yep, forgot to clean that out. It's been removed

dhlevi:
Any further debate on the `sourceSystemRef: 'mem-admin'` inclusion?

NickPhura:
> 
> 
> Any further debate on the `sourceSystemRef: 'mem-admin'` inclusion?

I think it comes down to whether or not that fields gets updated every time someone touches the record.

If it does, then updating it here makes sense.
If not, then it should be left as 'core', or w/e it was prior to this migration.

My assumption was the latter, but only because I was thinking of that field as more of a historical tag, so someone would know which source system to look in if they wanted to track down the original record.  But I think Mark was saying it was more akin to the 'updatedBy' field, where it just tracks the most recent source that updated the record. 

So, not really sure which is better, or if there are other pros/cons I'm not considering.

dhlevi:
It's kind of a mix... it holds historical reference to the source system. It's just in this case, we have two historical systems for one record. Original data will come from core, but it will be updated with source information from mem-admin. So, in this case I think we can assume a ref of `mem-admin` effectively represents Core+mem-admin, rather than a totally unique source ref.

By updating the ref, it also allows us to easily identify which records are touched by the BCMI migration in the event we need to re-run, roll back, etc.

NickPhura:
> By updating the ref, it also allows us to easily identify which records are touched by the BCMI migration in the event we need to re-run, roll back, etc.

Absolutely.  The sourceSystemRef will get updated to 'nrpti' I believe, if someone manually touches the record, but in the immediate moments after a migration, if we had to back it out you would still be able to, which is good.

api/migrations/20200603153526-bcmiDataLoad.js
```diff
@@ -0,0 +1,135 @@
...
+    for(const idx in mineData.externalLinks) {
+      const link = mineData.externalLinks[idx];
+      externalLinks.push(link.link);
+    }
+
+    return nrpti.update({ _id: new ObjectID(nrptiMines[0]._id) }, {
+      $set: {
+        type:            mineData.type,
+        summary:         mineData.description, // BCMI doesn't have a "summary" attribute
+        description:     mineData.description,
```
NickPhura, line 98: Not a PR issue, just some context:

The reason we have summary and description in some places, is because LNG uses description, but NRCED uses summary.  The reason being, is that NRCED internally uses both description and summary, but not in the same way LNG does.  So we ended up supporting both, depending on which flavour you're looking at. 

So if Mines also use both, then this is totally fine.  If not, then we'll probably end up removing one of them later on as we refine things.

dhlevi: Alright, good to know. I'm expecting this might be extended for document support in the short term as well, so we can readdress the model as needed.

PR #404, opened 2020-06-09
NRPT-71: Description & Summary keyword search.

NickPhura:
https://bcmines.atlassian.net/browse/NRPT-71
- Add description & summary option to admin keyword search
- Add description subset
- Add new description_summary_subset text index migration

Including just the description & summary part of this ticket.  
The final piece in the ticket is to add the recordName search, which I will add in a separate PR, as long as this one is on the right track.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.html
+++ b/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.html
@@ -40,6 +40,19 @@
                   check_circle
                 </i>
               </button>
+              <button
+                ngbDropdownItem
+                [disabled]="selectedSubset === 'Description & Summary'"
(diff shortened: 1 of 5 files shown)


NickPhura:
> 
> 
> LGTM except that you need to await before dropIndex/createIndex to be proper.

Updated: added the awaits

NickPhura:
> 
> 
> 👍 Looks good.
> 
> Do we used the array of Schema Names elsewhere (from the $match _schemaName $in)? If so, might be nice to put this in a constants file so we're sharing it rather then copy-pasting elsewhere that it's needed.

I think the only place where a list of schema names is used in this way, is in the subset code.  But we could add some constants to hold them.  I had sort of started a similar idea with the `record-type-enum.js`, but its not fully used everywhere at the moment.

PR #405, opened 2020-06-09
NRPTI-59: Remove Fee Orders

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-59

The criteria used to remove the fee orders comes from the ticket and some internal discussions. While the Fee Order should be of type 'Order' there is some bad data coming from Epic where they are also coming in as inspections, which is why the filtering happens on both types.


api/migrations/20200608243419-removeFeeOrders.js
```diff
@@ -0,0 +1,76 @@
...
+
+      // Delete any document records.
+      for (const documentId of order.documents) {
+        nrptiCollection.remove({ _id: documentId });
+      }
+
+      nrptiCollection.remove({ _id: order._id })
+    }
+    
+    mClient.close()
```
dhlevi, line 63: Minor: You could move mClient.close() out of the try and catch, and move it below (line 69/70) as it'll always be called regardless of the process outcome.

api/src/integrations/epic/orders-utils.js
```diff
@@ -62,6 +62,41 @@ class Orders extends BaseRecordUtils {
...
   }
+
+  /**
+   * Indicates if a record is a fee order or not.
+   * 
+   * @param {object} transformedRecord Epic record that has been transformed to NRPTI format
+   * @returns {boolean} Indication if the record is a fee order
+   * @memberof Orders
+   */
+  isRecordFeeOrder(transformedRecord) {
```
NickPhura, line 73: Does this also need to be added to the inspections utils, if the fee orders sometimes come in from epic as inspections?

api/src/integrations/epic/datasource.js
```diff
@@ -152,14 +152,26 @@ class DataSource {
...
+                processRecords.push(theRecord);
+              }
           }
           // Skip everything else
         } else {
-          processRecords.push(theRecord);
+          const rec = await recordTypeUtils.transformRecord(theRecord);
+
+          // Check if Order or Inspection and not a Fee Order 
+          if (rec._schemaName === 'Inspection' || rec._schemaName === 'Order') {
```
dhlevi, line 167: You have the condition:
`if (rec._schemaName === 'Inspection' || rec._schemaName === 'Order')`
Before calls to `recordTypeUtils.isRecordFeeOrder()`

Would it make more sense to push the condition into isRecordFeeOrder so you can remove the nested "Ifs"

api/migrations/20200608243419-removeFeeOrders.js
```diff
@@ -0,0 +1,76 @@
...
+
+    for (const order of feeOrders) {
+      // Delete any flavour records.
+      for (const flavourId of order._flavourRecords) {
+        nrptiCollection.remove({ _id: flavourId });
+      }
+
+      // Delete any document records.
+      for (const documentId of order.documents) {
+        nrptiCollection.remove({ _id: documentId });
```
NickPhura, line 57: We should also check for S3 documents

pseudo code:
```
const document = await Document.findOneAndRemove(...)
if(document.key)
then  DocumentController.deleteS3Document(key)
```

api/src/integrations/epic/orders-utils.js
```diff
@@ -62,6 +62,41 @@ class Orders extends BaseRecordUtils {
...
+    if (!transformedRecord) {
+      throw new Error('isRecordFeeOrder - required transformedRecord must be non-null.');
+    }
+
+    // Only want to handle Orders.
+    if (transformedRecord.recordType !== 'Order' || !transformedRecord.recordName) {
```
dhlevi, line 79: Logic question:

I would think `transformedRecord.recordType !== 'Order'` would result in a "false" from isRecordFeeOrder rather then throw an error? It's not necessarily an error case, just definitely not a fee order.

api/migrations/20200608243419-removeFeeOrders.js
KitArmstrong: I checked an it looks to me that all Epic documents are linked back to Epic's own URLs and we are not hosting any. Is that correct or do we indeed have some?

api/migrations/20200608243419-removeFeeOrders.js
KitArmstrong: This was consistent with the other migrations. I think I will opt to keep it matching them for the time being. Although you are right, we could place it just after the `catch`.

api/migrations/20200608243419-removeFeeOrders.js
NickPhura: All of the records imported from Epic will only be link documents.
The only way you'd get s3 documents is if an admin manually added one.

So there probably just aren't any

api/src/integrations/epic/datasource.js
KitArmstrong: It actually does exist in the `isRecordFeeOrder` function as well. However, many different record types come through here, but that function only exists on the two that need it. So if something other than Inspection or Order where to arrive, they can not call 'isRecordFeeOrder' as it won't exist.

api/src/integrations/epic/orders-utils.js
KitArmstrong: Great catch! Yes, I forgot to add it to the Inspection. I think I might add it to the base record just so all Epic types have access should they need it. Really there should be no Fee Orders arriving no matter what type they have been categorized as.

api/src/integrations/epic/orders-utils.js
KitArmstrong: This is being rewritten to address some other points too, so I don't think it will apply anymore.

api/migrations/20200608243419-removeFeeOrders.js
KitArmstrong: Okay, will update to include that check just in case.

api/src/integrations/epic/datasource.js
KitArmstrong: Removed the check by moving where the `isRecordFeeOrder` function lives.

KitArmstrong:
@dhlevi and @NickPhura I pulled out the transform method into the base utils so that it can be used by all Epic record types. Although it should really only be Orders, the data isn't consistent and the Fee Order might be under a different type so this allows all Epic records to run the check.

It should address most of the comments that were left. I also added a check in the migration for S3 documents.

api/migrations/20200608243419-removeFeeOrders.js
```diff
@@ -0,0 +1,83 @@
...
+        nrptiCollection.remove({ _id: flavourId });
+      }
+
+      // Delete any document records.
+      for (const documentId of order.documents) {
+        const doc = await nrptiCollection.findOneAndDelete({ _id: documentId });
+
+        // Delete from S3 if it exists.
+        if (doc.key) {
+          DocumentController.deleteS3Document(doc.key);
```
NickPhura, line 63: This probably needs an 'await'

api/src/integrations/epic/base-record-utils.js
```diff
@@ -231,6 +231,36 @@ class BaseRecordUtils {
...
+
+  /**
+   * Indicates if a record is a fee order or not.
+   * 
+   * @param {object} transformedRecord Epic record that has been transformed to NRPTI format
+   * @returns {boolean} Indication if the record is a fee order
+   * @memberof BaseRecordUtils
+   */
+  isRecordFeeOrder(transformedRecord) {
+    if (!transformedRecord || !transformedRecord.recordName) {
```
NickPhura, line 243: More of an epic data question: do we know for sure that all records will have a recordName? Or is it ever acceptable to not have a name (in which case we shouldn't throw an error here)?

api/migrations/20200608243419-removeFeeOrders.js
KitArmstrong: Added.

api/src/integrations/epic/base-record-utils.js
KitArmstrong: It takes in a transformed record which guarantees that there is a recordName property. If you feed it a non-transformed record it will not have this property and that is when it should error.

NickPhura: ```!transformedRecord.recordName``` will trigger on an empty string though, which is the default recordName if nothing comes from Epic

KitArmstrong: Good point. Updated to look for the actual object property rather than a falsey value.

PR #406, opened 2020-06-09
NPRT-42: Explicit ellipsis for public doc rows

dhlevi:
Small update to NRPT-42 for the NRCED public site to inline all document rows instead of wrapping text, add ellipsis on overflows.

### DIFF
--- a/angular/projects/public-nrpti/src/app/records/administrative-penalties/administrative-penalty-detail/administrative-penalty-detail.component.html
+++ b/angular/projects/public-nrpti/src/app/records/administrative-penalties/administrative-penalty-detail/administrative-penalty-detail.component.html
@@ -96,7 +96,7 @@
           <div class="grid-item__row">
             <label class="grid-item-name">Documents</label>
             <span class="grid-item-value__col" *ngIf="data && data.documents && data.documents.length > 0">
-              <span *ngFor="let document of data && data.documents">
+              <span class="trim-document-text"  *ngFor="let document of data && data.documents">
                 <a href="{{ document.url }}" target=" _blank">{{ document.fileName }}</a>
               </span>
             </span>
--- a/angular/projects/public-nrpti/src/app/records/administrative-sanctions/administrative-sanction-detail/administrative-sanction-detail.component.html
+++ b/angular/projects/public-nrpti/src/app/records/administrative-sanctions/administrative-sanction-detail/administrative-sanction-detail.component.html
@@ -92,7 +92,7 @@
           <div class="grid-item__row">
             <label class="grid-item-name">Documents</label>
             <span class="grid-item-value__col" *ngIf="data && data.documents && data.documents.length > 0">
-              <span *ngFor="let document of data && data.documents">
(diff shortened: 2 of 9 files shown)


angular/projects/public-nrpti/src/app/records/administrative-penalties/administrative-penalty-detail/administrative-penalty-detail.component.html
```diff
@@ -96,7 +96,7 @@
           <div class="grid-item__row">
             <label class="grid-item-name">Documents</label>
             <span class="grid-item-value__col" *ngIf="data && data.documents && data.documents.length > 0">
-              <span *ngFor="let document of data && data.documents">
+              <span class="trim-document-text"  *ngFor="let document of data && data.documents">
                 <a href="{{ document.url }}" target=" _blank">{{ document.fileName }}</a>
               </span>
```
NickPhura, line 101: This is maybe not what this ticket was aiming to address, but something you can add if you wanted to.  Its been in the back of my mind for a while, and this sparked my memory.

If there are multiple documents, they render in a row, rather than in columns.

This would be fixed by changing:

```
            <span class="grid-item-value__col" *ngIf="data && data.documents && data.documents.length > 0">
              <span class="trim-document-text"  *ngFor="let document of data && data.documents">
                <a href="{{ document.url }}" target=" _blank">{{ document.fileName }}</a>
              </span>
            </span>
```

to

```
            <span class="grid-item-value" *ngIf="data && data.documents && data.documents.length > 0">
              <ng-container *ngFor="let document of data && data.documents">
                <span class="trim-document-text">
                  <a href="{{ document.url }}" target=" _blank">{{ document.fileName }}</a>
                </span>
              </ng-container>
            </span>
```

IE: changing ```class="grid-item-value__col"``` to ```class="grid-item-value"``` and moving the ```*ngFor="let document of data && data.documents"``` into a wrapping ```<ng-container>```.

dhlevi: That's a good idea actually. It'll help keep the row truncating sensible as well... previously it wrapped but reviewing the code, now you could get merged lines.

NickPhura: Ill give this the thumbs up for now.
If you do end up adding this change, then Ill take another look.  If not, then whats here looks good to go.

dhlevi: I decided to include this change into this update. Makes sense and it'll keep the document links looking much better. No sense in making a second ticket for this.

NickPhura: Sweet! Looks good! 👍

PR #409, opened 2020-06-10
NRPT-69: Added filter via publish status of flavours to admin

danieltruong:


NickPhura:
I noticed an issue, that came up while testing this (though I don't think it was caused by this PR).

Steps that led to finding the issue:
- Wipe database
- Run migrations
- Choose an Order that had been published to NRCED, but which did not have an LNG flavour created yet.
- Edit the LNG flavour, and toggle the publish to true.
- Save record.

Tried to find the record using the publish filters:
- Selected `NrcedPublished=True` and `LngPublished=True`
- No results found.

Looking at the database:
- The master record did not have its `isLngPublished` set to true, it was still false.
- Looking at the flavour records themselves, I noticed that the _master backref for the NRCED flavour was correctly pointing to the master record.  But the _master backref for the LNG flavour was not.  It had an ObjectId for a record that didn't exist.

So I think something weird is happening with the _master backref, which is then causing the publish filter to fail, since it uses the backref when updating the `is<Lng/Nrced>Published` fields.

NickPhura:
I tried creating a new record, with both flavours edited and published, and it worked as expected.  Both get the correct _master backref, and both cause the `is<Lng/Nrced>Published` field to update correctly.


I tried creating a new record, with just 1 flavour edited and published, and I get an error:
```Cannot read property 'addRole' of undefined```

I suspect the the create/edit RecordWithFlavour function needs some extra null checking.

danieltruong:
> See comments

2719623f175bd5f8239b264be52a033213168752

NickPhura:
Pulled down the latest changes.

I'm still seeing some issues with the _mater backref, when a flavour record is created for an existing record.

It only seems to happen when a flavour is created during a Put.  The flavour gets an incorrect _master backref, and then the other parts that use _master don't work as expected.

danieltruong:
> Pulled down the latest changes.
> 
> I'm still seeing some issues with the _mater backref, when a flavour record is created for an existing record.
> 
> It only seems to happen when a flavour is created during a Put. The flavour gets an incorrect _master backref, and then the other parts that use _master don't work as expected.

5a08897162e84c91b4a3a53188d7912eee2d8d64

NickPhura:
> 
> 
> > Pulled down the latest changes.
> > I'm still seeing some issues with the _mater backref, when a flavour record is created for an existing record.
> > It only seems to happen when a flavour is created during a Put. The flavour gets an incorrect _master backref, and then the other parts that use _master don't work as expected.
> 
> [5a08897](https://github.com/bcgov/NRPTI/commit/5a08897162e84c91b4a3a53188d7912eee2d8d64)

Looks like it is working as expected now

I think its important that we write some unit tests for these 2 functions though.  In particular, the main loops that do all of the array `entry` logic.

danieltruong:
> > > Pulled down the latest changes.
> > > I'm still seeing some issues with the _mater backref, when a flavour record is created for an existing record.
> > > It only seems to happen when a flavour is created during a Put. The flavour gets an incorrect _master backref, and then the other parts that use _master don't work as expected.
> > 
> > 
> > [5a08897](https://github.com/bcgov/NRPTI/commit/5a08897162e84c91b4a3a53188d7912eee2d8d64)
> 
> Looks like it is working as expected now
> 
> I think its important that we write some unit tests for these 2 functions though. In particular, the main loops that do all of the array `entry` logic.

47ddfb3

PR #410, opened 2020-06-11
NRPT-71-2: recordName keyword search

NickPhura:
https://bcmines.atlassian.net/browse/NRPT-71

- add admin recordName subset keyword search
- add recordName subset
- add recordName subset text index migration

As mentioned in [PR-404](https://github.com/bcgov/NRPTI/pull/404) this is the final remaining piece of the keyword search ticket.


api/materialized_views/search/recordNameSubset.js
```diff
@@ -0,0 +1,95 @@
...
+                'WarningLNG'
+              ]
+            ]
+          },
+          then: '$$KEEP',
+          else: '$$PRUNE'
+        }
+      }
+    }
+  ];
```
marklise, line 80:
```
await collection.createIndex(
  {
    'flavours.recordName': 'text'
  },
  {
    name: 'record-name-subset-text-index'
  }
);
```
This is ultimately asking to index against the `recordName` property on LNG records only.  It's very inefficient to query the master records, populate everything, then redact, when you could have matched against LNG records only, insert them into the collection, and then create an index on that field.  Records that don't have recordName wouldn't be returned in the result because the index wouldn't hit, or your API call against this indexed collection could query of $text _and_ _schemaName: 'xxx'.

NickPhura: I think we could use this [partial index concept](https://docs.mongodb.com/manual/core/index-partial/) to create text indexes on only LNG schemas.  

And then this subset could be the same as the description_summary one, where we only do a lookup on flavours, and thats it.

NickPhura: @marklise I've updated this PR as per Jess's comment (https://bcmines.atlassian.net/browse/NRPT-71).  So it is now just a regular subset on a master field.

NickPhura:
> 
> 
> I think we need to move the recordName UP to the master record right now, as part of this ticket. The flavour change logic can be done in a separate ticket.

Moved the record name field up in the add-edit and details components.

PR #414, opened 2020-06-11
NRPT-163: Updates Core importer mapping

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-163

Updates the Core importer to map the status and commodities properly.

### DIFF
--- a/api/src/integrations/core/mine-utils.js
+++ b/api/src/integrations/core/mine-utils.js
@@ -79,11 +79,13 @@ class Mines extends BaseRecordUtils {
     const commodities = [];
 
     for (const mineType of mineRecord.mine_type) {
-      for (const typeDetail of mineType.mine_type_detail) {
-        // Could be null if type is a disturbance instead of a commodity.
-        if (typeDetail.mine_commodity_code) {
-          const commodity = commodityTypes.find(commodity => commodity.mine_commodity_code === typeDetail.mine_commodity_code);
-          commodities.push(commodity.description);
+      if (mineType.mine_type_detail && mineType.mine_type_detail.length) {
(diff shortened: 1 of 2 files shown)


api/src/integrations/core/mine-utils.js
```diff
@@ -101,8 +103,13 @@ class Mines extends BaseRecordUtils {
       throw new Error('getLatestStatus - mineRecord must not be null.');
     }
 
-    const latestStatus = mineRecord.mine_status.pop();
-    return (latestStatus && latestStatus.status_labels.join(' ')) || '';
+    if (!mineRecord.mine_status.length) {
```
NickPhura, line 106: Do we need a null check for mine_status? Or should it always be non-null?

api/src/integrations/core/mine-utils.js
```diff
@@ -78,13 +78,17 @@ class Mines extends BaseRecordUtils {
 
     const commodities = [];
 
-    mineRecord.mine_type.forEach(type => {
-      commodityTypes.forEach(commodity => {
-        if (commodity.mine_tenure_type_codes.includes(type.mine_tenure_type_code)) {
-          commodities.push(commodity.description);
+    for (const mineType of mineRecord.mine_type) {
```
KitArmstrong, line 81: `mine_type` is checked above so that a proper error is thrown if it is missing.

api/src/integrations/core/mine-utils.js
KitArmstrong: Updated

KitArmstrong:
I fixed a unit test and added in a few additional null checks. The data coming from Core API has null values, but the properties on the objects always exist. I still added the checks in case that ever changes though.

PR #417, opened 2020-06-12
NRPT-88: Schedule for epic import

marklise:


api/src/tasks/import-task.js
```diff
@@ -93,18 +115,10 @@ function getDataSourceConfig(dataSourceType) {
...
-        dataSourceClass: require('../integrations/nris/datasource')
-      };
-    default:
-      return null;
-  }
+  // dataSourceType will match the name of a directory for the given
+  // integration in /src/integrations/<dataSourceType>/
+  return {
+    dataSourceLabel: dataSourceType,
+    dataSourceClass: require(`../integrations/${dataSourceType}/datasource`)
```
NickPhura, line 122: @marklise I don't think this works for `nris` currently, as the folder structure is `integrations/nris/stuff` but all of our calls, and swagger, use 'nris-epd' as the dataSourceType.

PR #420, opened 2020-06-16
NRPT-138: Adds new roles

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-138

I updated the Swagger to include the new roles which will prevent any API calls from proceeding. I also added a check in the corresponding function just in case it were to every be called without Swagger. 

Right now most calls allow any admin user as they all go through the same endpoints. Where there was specific logic due to a certain flavour I have added that restriction.

### DIFF
--- a/api/src/controllers/document-controller.js
+++ b/api/src/controllers/document-controller.js
@@ -1,11 +1,14 @@
 const mongoose = require('mongoose');
+const AWS = require('aws-sdk');
+const ObjectID = require('mongodb').ObjectID;
+
 const queryActions = require('../utils/query-actions');
 const queryUtils = require('../utils/query-utils');
 const businessLogicManager = require('../utils/business-logic-manager');
-const AWS = require('aws-sdk');
 const mongodb = require('../utils/mongodb');
-const ObjectID = require('mongodb').ObjectID;
-let defaultLog = require('../utils/logger')('record');
+const defaultLog = require('../utils/logger')('record');
+const { userInRole } = require('../utils/auth-utils');
+const { ROLES } = require('../utils/constants/misc');
 
 const OBJ_STORE_URL = process.env.OBJECT_STORE_endpoint_url || 'nrs.objectstore.gov.bc.ca';
 const ep = new AWS.Endpoint(OBJ_STORE_URL);
@@ -21,7 +24,11 @@ exports.protectedOptions = function(args, res, next) {
   res.status(200).send();
 };
 
-exports.protectedPost = async function(args, res, next) {
+exports.protectedPost = async function(args, res, next) { // Confirm ser has correct role.
+  if (!userInRole(ROLES.ADMIN_ROLES, args.swagger.params.auth_payload.realm_access.roles)) {
+    return queryActions.sendResponse(res, 400, 'Missing valid user role.');
+  }  
+
   if (
     args.swagger.params.fileName &&
     args.swagger.params.fileName.value &&
@@ -136,6 +143,10 @@ exports.protectedPost = async function(args, res, next) {
 };
 
 exports.protectedDelete = async function(args, res, next) {
+  if (!userInRole(ROLES.ADMIN_ROLES, args.swagger.params.auth_payload.realm_access.roles)) {
+    return queryActions.sendResponse(res, 400, 'Missing valid user role.');
+  }
+
   if (
     args.swagger.params.docId &&
     args.swagger.params.docId.value &&
@@ -442,6 +453,10 @@ exports.unpublishDocument = unpublishDocument;
  * @returns
  */
 exports.protectedGetS3SignedURL = async function(args, res, next) {
+  if (!userInRole(ROLES.ADMIN_ROLES, args.swagger.params.auth_payload.realm_access.roles)) {
+    return queryActions.sendResponse(res, 400, 'Missing valid user role.');
+  }  
+  
   if (!args.swagger.params.docId || !args.swagger.params.docId.value) {
     defaultLog.warn('protectedGet - missing required docId param');
     return queryActions.sendResponse(res, 400, 'Missing required docId param');
--- a/api/src/controllers/document-controller.js
+++ b/api/src/controllers/document-controller.js
@@ -24,7 +24,7 @@ exports.protectedOptions = function(args, res, next) {
   res.status(200).send();
 };
 
-exports.protectedPost = async function(args, res, next) { // Confirm ser has correct role.
+exports.protectedPost = async function(args, res, next) { // Confirm user has correct role.
   if (!userInRole(ROLES.ADMIN_ROLES, args.swagger.params.auth_payload.realm_access.roles)) {
     return queryActions.sendResponse(res, 400, 'Missing valid user role.');
   }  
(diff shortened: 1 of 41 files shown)


api/src/utils/auth-utils.js
```diff
@@ -150,3 +150,29 @@ function verifySecret(currentScopes, tokenString, secret, req, callback, sendErr
...
 }
+
+/**
+ * Checks if a user has a role against an array of valid roles.
+ * 
+ * @param {Array<string>|string} validRoles Roles to search for a match.
+ * @param {Array<string>|string} userRoles Roles to match against.
+ * @returns {boolean} Indication if a match is found.
+ */
+exports.userInRole = function(validRoles, userRoles) {
```
dhlevi, line 161: Petty little quibble. Rename function to `userInRoles` as it accepts an array of validRoles, vs. a single role to check.

api/src/swagger/swagger.yaml
```diff
@@ -731,37 +750,6 @@ paths:
           description: 'Access Denied'
           schema:
             $ref: '#/definitions/Error'
-    put:
```
dhlevi, line 734: Nice bit of cleanup 👍

api/src/controllers/document-controller.js
```diff
@@ -1,11 +1,14 @@
...
+
 const queryActions = require('../utils/query-actions');
 const queryUtils = require('../utils/query-utils');
 const businessLogicManager = require('../utils/business-logic-manager');
-const AWS = require('aws-sdk');
 const mongodb = require('../utils/mongodb');
-const ObjectID = require('mongodb').ObjectID;
-let defaultLog = require('../utils/logger')('record');
+const defaultLog = require('../utils/logger')('record');
+const { userInRole } = require('../utils/auth-utils');
```
marklise, line 10: None of these checks for userInRole are required.  Swagger security takes care of that for us by specifying the roles allowed to even make the API call in the first place.  The pattern is to reject the API call as early as possible, so as not to have your API DOS'ed.  This is done at token decryption time in the auth-utils.js file, function verifySecret():

```
      if (roleMatch && issuerMatch) {
        // add the token to the request so that we can access it in the endpoint code if necessary
        req.swagger.params.auth_payload = decodedToken;
        defaultLog.info('JWT Verified');
        return callback(null);
      } else {
        defaultLog.info('JWT Role/Issuer mismatch');
        return callback(sendError());
      }
```

Only if a user has a scope defined at the API level in swagger will it then call the protectedXXXX function.  Here's the definition for protectedPut:

```
      security:
        - Bearer: []
      x-security-scopes:
        - sysadmin
        - admin:lng
        - admin:nrced
        - admin:bcmi
```

This means that anyone who has any one of these roles will be forwarded onto the protectedXXXX function.

The only thing you might do (which you're currently not doing) is specify a _general_ set of high level scopes at the API level to prevent unnecessary API calls (as you have done).  Then only allow the change to the specific record if that user has a matching role inside the `write` property of the document (document snipped for brevity):

```
{
    "_id" : ObjectId("5ee3fd432cdf09458efb7ca8"),
    ...
    "_schemaName" : "OrderLNG",
    ...
    "read" : [ 
        "sysadmin", 
        "public"
    ],
    "write" : [ 
        "sysadmin"
    ],
    "recordName" : "Section 34 Order - 2019-01-09 - Bear Fencing Rescinded ",
    ...
}
```

In the record above, you would want to create a mongo update to the object only if the user's scope match one of the entries inside the `write` property, in this case, it's sysadmin only.  There are a number of ways to do this, but ideally it's part of the update call which mongo should reject as part of the query.  NB: **Not** the following trivial node-js approach: query find -> get result -> compare array write properties -> if ok then mongo.update().. since that's not an atomic operation.  The record may have changed since that piece of code 'found' the record and then you blindly let the update happen.

In adding roles to the system using the record example above, we also need to consider adding the roles to people in KeyCloak, and then making a migration task to update the desired set of records according to who can write. ie: add `admin:lng` to the `write` property of _schemaName=OrderLNG,InspectionLNG...****LNG etc.. and the same for BCMI as well as NRCED.  Here's an updated record as an example:

```
{
    "_id" : ObjectId("5ee3fd432cdf09458efb7ca8"),
    ...
    "_schemaName" : "OrderLNG",
    ...
    "read" : [ 
        "sysadmin",
        "admin:lng",
        "public"
    ],
    "write" : [ 
        "sysadmin",
        "admin:lng" 
    ],
    "recordName" : "Section 34 Order - 2019-01-09 - Bear Fencing Rescinded ",
    ...
}
```

Master records will need some different thoughts as well.  In those cases, the read might be all the admin:* roles, but maybe the flavours get all write (since we have some properties that change the master across the board).

dhlevi: There are tickets already created for adding scopes to users as well as a data migration for updating docs in mongodb.

So, to be clear, the update here should be checking if the doc and user contain the matching scope in `write`, as the swagger middleware has verified the token already. We know the user has the scope, or we shouldn't be here... so for the flavours:

`db.collection.updateOne({ _id: '...', write: { $all: [scopes] }}, myDoc)...`

And creates need to add the scopes to `write` as needed.

Masters are admin:* editable, so we just need to verify the token via swagger, which Kit already added in the swagger doc.

KitArmstrong: I have already added all of the roles to the Swagger, but that doesn't restrict the ability to perform actions on certain flavours due to all types of records/documents using the same endpoint. Most of the checks with the `userInRole` prevent someone who has made it past the Swagger check, but should not be altering a certain flavour.

Adding in a check to look at the `write` property would accomplish this as well, so I will make the adjustments to take that approach.

As Dylan noted, migrations and user roles are all covered in different tickets.

marklise: @KitArmstrong @dhlevi re: Migrations, sorry I wasn't aware of the other tix when I was reviewing, just adding my thoughts around what was likely needed.  Sounds like you got it covered 👍 

@dhlevi correct, but a small tweak on the `$all`. I think that should probably be a `$in` or some other conditional.  The rule is if the caller of the api has any _one of the_ scopes defined inside the `write` property, they get write privileges on that particular document.  It doesn't have to be a perfect match.

api/src/controllers/post/administrative-penalty.js
```diff
@@ -198,6 +200,11 @@ exports.createMaster = function (args, res, next, incomingObj, flavourIds) {
  * @returns created lng administrativePenalty record
  */
 exports.createLNG = function (args, res, next, incomingObj) {
+  // Confirm user has correct role to create this type of record.
```
KitArmstrong, line 203: This check still has to exist for all `create[flavour]` calls in any Post controllers.  We need to determine that a user has permission to create this type of flavour record. We cannot check the master record as all admins have permission to write on those.

KitArmstrong:
I have moved most of the role verifications to the actual DB calls. This removes the check from all of the Put controllers; however, a few manual checks still remain in the Post controllers. We need to check that a user has permission to create a new flavour record and we cannot look at the `write` array on the master record because all admins have permission to the master record. So, the check still remains only for the flavour records in the Post controllers.

PR #421, opened 2020-06-16
NRPT-66: UI Tweaks

danieltruong:


danieltruong:
> 👍 Looks good.
> 
> Definitely not a breaker, but I noticed that the labels for the radio buttons for "Published/Unpublished" squish together on smaller widths. Not sure if we want to stack them on top of each other similar to the "Entity Type" checkboxes instead?

3c1178e

PR #422, opened 2020-06-16
NRPT-72: Enabled multiple source systems for filtering.

danieltruong:


angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
```diff
@@ -207,7 +207,7 @@ export class RecordsListComponent implements OnInit, OnDestroy {
       act: new FormControl((this.queryParams && this.queryParams.act) || null),
       regulation: new FormControl((this.queryParams && this.queryParams.regulation) || null),
       activityType: new FormControl((this.queryParams && this.queryParams.activityType) || null),
-      sourceSystemRef: new FormControl((this.queryParams && this.queryParams.sourceSystemRef) || null),
+      sourceSystemRef: new FormControl((this.queryParams && this.queryParams.sourceSystemRef.split(',')) || null),
```
dhlevi, line 210: This line fails if sourceSystemRef is null. Probably needs a null check.

angular/projects/admin-nrpti/src/app/explore-panel/explore-panel.component.html
```diff
@@ -147,7 +148,7 @@
         </div>
       </div>
       <div class="col-4">
-        <label for="sourceSystemFilter">Source System</label>
+        <!-- <label for="sourceSystemFilter">Source System</label>
```
dhlevi, line 151: Probably can remove the commented out stuff, uness it's being kept for re-adding or reference?

angular/projects/admin-nrpti/src/app/explore-panel/explore-panel.component.html
```diff
@@ -158,7 +159,11 @@
           <option [selected]="selectedSystemRef === 'ocers-csv'" value="ocers-csv">OCERS</option>
           <option [selected]="selectedSystemRef === 'nris-epd'" value="nris-epd">NRIS</option>
           <option [selected]="selectedSystemRef === 'epic'" value="epic">EPIC</option>
-        </select>
+        </select> -->
+        <mat-label>Source System</mat-label>
```
dhlevi, line 163: CSS styling is different from the other UI components on the screen, looks like it's just because it's a select vs. a auto-complete select like ActivityType, etc. Did we want this component to work the same as the other selectors in the filter panel?

angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
marklise: > This line fails if sourceSystemRef is null. Probably needs a null check.

I'm getting the following on load of the list records page:
image

angular/projects/admin-nrpti/src/app/explore-panel/explore-panel.component.html
danieltruong: The select has different functionality as auto-complete. At the moment app-autocomplete-multi-select has a few bugs as well. It is likely we will re-write it. During that time, we can unify the styling between select and auto-complete.

angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
danieltruong: bafd6120d732c2dc459f782faee36eaa19dbd513

angular/projects/admin-nrpti/src/app/explore-panel/explore-panel.component.html
danieltruong: bafd6120d732c2dc459f782faee36eaa19dbd513

PR #424, opened 2020-06-17
NRPT-136: Audit logging update

dhlevi:
Update to Audit Log to store IDIR uuid, preferred and display name as a JSON string.

Adding additional audit log calls to record and document POST, PUT, and DELETE's. Adding Audits to document record and flavour updates on document changes (Not sure if we need them, so these can be remove if unecessary).

Removed the audit call in schema.save() because that was generating empty audit records, and would require some extensive reworking to inject with meta, actions, and proper user info that's beyond the scope of this ticket.

See ticket [NRPT-136](https://bcmines.atlassian.net/browse/NRPT-136)


dhlevi:
Apologies for the commit history. I did a squash with Fork (the git UI tool) rather than via the git command line... apparently it doesn't work as expected...

api/src/controllers/document-controller.js
```diff
@@ -54,6 +54,8 @@ exports.protectedPost = async function(args, res, next) { // Confirm user has co
           args.swagger.params.url.value,
           readRoles
         );
+
+        await queryUtils.recordAction('POST', JSON.stringify(docResponse), args.swagger.params.auth_payload, docResponse.key);
```
NickPhura, line 58: Its not super obvious, but this `docResponse` will never have a `key`.

The `key` is the S3 key, so it only gets created if you create a document via `createS3Document()`

NickPhura: Just noticed: I think that parameter wants the object _id, so would want to pass `docResponse._id`.  I think this applies to all of the recordActions where `docResponse.key` is being set.

api/src/controllers/record-controller.js
```diff
@@ -169,6 +169,13 @@ exports.protectedPost = async function (args, res, next) {
 
     let response = await Promise.all(promises);
 
+    // Audit the POST action.
+    // If multiple observables are triggered, response will have multiple objects
+    // only add the metaId for single objects.
+    let meta = response && response[0] && response[0][0] ? response[0][0] : null;
+    let metaID = meta ? meta.object[0]._id : null;
```
NickPhura, line 176: Does this require null checking similar to the line above (175)?

api/src/utils/query-utils.js
```diff
@@ -72,14 +72,18 @@ exports.getSkipLimitParameters = function(pageSize, pageNum) {
...
   const audit = new Audit({
     _objectSchema: 'Query',
     action: action,
     meta: meta,
     objId: objId,
-    performedBy: username
+    performedBy: JSON.stringify({
+      idir_userid: authPayload.idir_userid,
+      displayName: authPayload.displayName,
+      preferred_username: authPayload.preferred_username
```
NickPhura, line 85: Might need to null check authPayload

NickPhura: Could even default these values to something that indicates the auth_payload was null, which presumably isn't desirable

api/src/controllers/document-controller.js
dhlevi: Sounds good. I'll update those now.

api/src/controllers/record-controller.js
dhlevi: Yes. Check has been added to ensure object is there

api/src/utils/query-utils.js
dhlevi: Added a check (and a try/catch for extra padding)

dhlevi:
I've made a bit of a change to the implementation here. Nothing to crazy, but certainly means we'll need another look just in case.

Changes are:

Addition of a middleware for handling Audit calls after the request has been completed
Changing all Audit calls to use a new audit function in `queryUtils`
Removing the `return` statements in the endpoint controllers and using `next()` to trigger the middleware.

PR #427, opened 2020-06-17
NRPT-137: Migration to update roles in all records

dhlevi:
Migration script that will update all records in NRPTI to contain their appropriate read/write roles.

Master documents `read` array will contain all roles, plus `public` where it exists today
Flavour documents `read` array will contain all roles, plus `public` where it exists today

Master documents `write` array will contain all admin roles.
Flavour documents `write` array will contain flavour specific role (`sysadmin` and either `admin:lng` or `admin:nrpti`) only.

See ticket [NRPT-137](https://bcmines.atlassian.net/browse/NRPT-137)


api/migrations/20200617203111-roleUpdates.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+exports._meta = {
+  "version": 1
+};
+
+async function documentUpdates(nrpti) {
+  let promises = [];
+
+  const masterUpdate = { $set: { write: ALL_ROLES }, $push: { read: { $each: [roles.ADMIN_LNG, roles.ADMIN_NRCED, roles.ADMIN_BCMI] }}};
```
KitArmstrong, line 89: Should the read array be explicitly set just in case there is some bad data in there or if sysadmin is missing? Right now it is just appending to whatever already exists.

dhlevi: Good catch. Changing to `addToSet`

dhlevi: Follow up note: The reason I'm not explicitly setting the read arrays is so I don't change any existing `public` roles. I could do that with a static set, but I'd have to check the array first, push/addToSet ensures we inject only the new roles without altering the existing roles, valid or otherwise. `$addToSet` is better then `$push` in this case because we don't want dupes on a re-run of the migration.

api/migrations/20200617203111-roleUpdates.js
```diff
@@ -0,0 +1,117 @@
...
+
+async function documentUpdates(nrpti) {
+  let promises = [];
+
+  const masterUpdate = { $set: { write: ALL_ROLES }, $addToSet: { read: { $each: ALL_ROLES }}};
+  const lngUpdate    = { $set: { write: [roles.ADMIN, roles.ADMIN_LNG] }, $addToSet: { read: { $each: ALL_ROLES }}};
+  const nrcedUpdate  = { $set: { write: [roles.ADMIN, roles.ADMIN_NRCED] }, $addToSet: { read: { $each: ALL_ROLES }}};
+  const options      = { "upsert": false };
+
+  for(const recordType in RECORD_TYPE) {
```
KitArmstrong, line 94: Nit: Careful with `for...in` over objects as they will go up the prototype chain if there are higher enumerable properties. Should be okay because this is just a migration.

dhlevi: Yeah, `for (let [key, value] of Object.entries(RECORD_TYPE))` would be safer and cleaner...

PR #428, opened 2020-06-18
NOBUG: Remove migration, not necessary anymore.

marklise:


NickPhura:
Can we really remove this migration? 
Currently, it is the only migration that imports `ocers-data-2.csv`.
How would that second ocers csv get imported if not through this migration?

marklise:
> Can we really remove this migration?
> Currently, it is the only migration that imports `ocers-data-2.csv`.
> How would that second ocers csv get imported if not through this migration?

Right now it's bombing out a clean migration run as it can't find the file.  I think it should be removed in order for no errors to exist in case someone clones the repo cleanly.  The other one 'ocers-data.csv' should also get tossed as that file doesn't exist in the repo anymore.  I was going to do that in a separate PR as it has already been run against prod.

PR #431, opened 2020-06-18
NRPT-186: Adding missing record types to activityTypePicklist

danieltruong:


angular/projects/common/src/app/utils/record-constants.ts
```diff
@@ -81,11 +81,17 @@ export class Picklists {
...
     AdministrativePenalty: { displayName: 'Administrative Penalty', _schemaName: 'AdministrativePenalty' },
     AdministrativeSanction: { displayName: 'Administrative Sanction', _schemaName: 'AdministrativeSanction' },
+    Agreement: { displayName: 'Agreement', _schemaName: 'Agreement' },
+    Certificate: { displayName: 'Certificate', _schemaName: 'Certificate' },
+    ConstructionPlan: { displayName: 'Construction Plan', _schemaName: 'ConstructionPlan' },
     CourtConviction: { displayName: 'Court Conviction', _schemaName: 'CourtConviction' },
     Inspection: { displayName: 'Inspection', _schemaName: 'Inspection' },
+    ManagementPlan: { displayName: 'Management Plan', _schemaName: 'ManagementPlan' },
     Order: { displayName: 'Order', _schemaName: 'Order' },
     RestorativeJustice: { displayName: 'Restorative Justice', _schemaName: 'RestorativeJustice' },
```
NickPhura, line 91: Missing `Permit`

danieltruong: 7f0799b37b8068e8d0e0698a0f98031542b7b081

angular/projects/common/src/app/utils/record-constants.ts
```diff
@@ -2,15 +2,15 @@ import { Utils } from './utils';
 import { Legislation } from '../models/master/common-models/legislation';
 
 export class EpicProjectIds {
-  public static readonly lngCanadaId = '588510cdaaecd9001b815f84';
+  public static readonly lngCanadaId = '588511d0aaecd9001b826192';
```
NickPhura, line 5: Accidental change?

marklise: See https://bcmines.atlassian.net/browse/NRPT-157.

NickPhura: Correct me if I followed that thread wrong: so this ID replaces the old one entirely? 
In that case, there are other instances of the old ID in the codebase that probably need replacing.  We would need a migration then as well to update the old IDs?

marklise: I believe that is coming soon

NickPhura: sweet 👍

PR #435, opened 2020-06-18
NRPT-106: Adding search to Mines

dhlevi:
Adding the search bar to the Mines list page, with no filter functionality. As per design, adding a "clear keywords" button. Additionally, adding a migration to update the keywords index for supporting Mines records permitNumber as part of the search.

Added optional bonus (can be removed if not needed) added a fuzzySearch nGram function (previously written for epic searches) to ensure partial word searches are scored more appropriately (brought it in for permit number searching, partial record name, etc, will only be active on `schemaName == 'Mine'`).

See Ticket [NRPT-106](https://bcmines.atlassian.net/browse/NRPT-106)


danieltruong:
keywordIndexUpate.js should be keywordIndexUpdate.js

dhlevi:
> keywordIndexUpate.js should be keywordIndexUpdate.js

Whoops. Fixed (along with the broken unit test).

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.html
```diff
@@ -1,4 +1,40 @@
 <main class="container-fluid-padding">
+  <h1>All Mines</h1>
+  <br />
+  <div class="border-bottom">
+    <div class="row mb-3">
+      <div class="col">
+        <span style="font-size: 1.25rem;">Keyword Search</span>
```
KitArmstrong, line 7: Can we remove the inline CSS?

api/src/controllers/search.js
```diff
@@ -205,6 +206,11 @@ let searchCollection = async function (
   // optional search keys
   let searchProperties = undefined;
   if (keywords) {
+    // for now, limit fuzzy search to the mine search only. We can expand to all searches
+    // later if desired
+    if (schemaName === 'Mine') {
```
KitArmstrong, line 211: This proved extremely useful in Epic, glad to see we can use it here too.

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.html
dhlevi: Yep. I just copied from the search in records, but I can change it in both places and move it into the CSS file.

api/migrations/20200618184040-keywordIndexUpdate.js
```diff
@@ -0,0 +1,54 @@
...
+      summary: 'text',
+      location: 'text',
+      'issuedTo.companyName': 'text',
+      'issuedTo.firstName': 'text',
+      'issuedTo.middleName': 'text',
+      'issuedTo.lastName': 'text',
+      'issuedTo.fullName': 'text',
+      permitNumbers: 'text',
+      recordName: 'text',
+      name: 'text'
```
danieltruong, line 36: Are these all the fields we want to index? The ticket doesn't seem to say what we should be indexing on. Would it be a good idea to add description, permittee, type, status and region? That way we can search on the columns that we see on the mines list.

dhlevi: The ticket only indicates Name and Permit Numbers. It's easy enough to add other fields if requested of course.

dhlevi:
> Type in something into the keyword search. Search. Refresh the page. They keywords are correctly populated in the URL but do not show up in the search box in the UI.

Fixed!

PR #438, opened 2020-06-19
NRPT-104: Updates Mines list page

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-104

Updates the mine list page. Note the comments in the ticket as there are a couple implementation differences from the mock ups.


angular/projects/admin-nrpti/src/app/mines/mines-rows/mines-table-row.component.html
```diff
@@ -1,8 +1,25 @@
...
+  <div *ngIf="rowData.publishedDate; then published else unpublished"></div>
+  <ng-template #published class="test">
+    <i class="material-icons published-green">check_circle</i>
+    <span class="publish-label pl-1 published-green">Published</span>
+  </ng-template>
+  <ng-template #unpublished class="test">
+    <i class="material-icons ">cancel</i>
+    <span class="publish-label pl-1">Unpublished</span>
+  </ng-template>
+</td>
```
danieltruong, line 22: We should probably base publish state off of whether 'public' exists in the read array. Published date my be populated if a mine is published and then unpublished.

KitArmstrong: Good point. Updated to look at the read array.

angular/projects/admin-nrpti/src/app/mines/mines-rows/mines-table-row.component.html
```diff
@@ -1,8 +1,25 @@
...
+<td scope="row" data-label="Type" class="col-1">{{ rowData.type || '-' }}</td>
 <td scope="row" data-label="Permittee" class="col-2">{{ rowData.permittee || '-' }}</td>
-<td scope="row" data-label="Type" class="col-2">{{ rowData.type || '-' }}</td>
-<td scope="row" data-label="Status" class="col-2">{{ rowData.status || '-' }}</td>
-<td scope="row" data-label="Region" class="col-2">{{ rowData.region || '-' }}</td>
+<td scope="row" data-label="Region" class="col-1">{{ rowData.region && displayRegion(rowData.region) || '-' }}</td>
+
+<td scope="row" data-label="Published State" class="col-2 d-flex">
+  <div *ngIf="isPublic(); then published else unpublished"></div>
```
danieltruong, line 13: Don't really need a function here. You could directly put in this.rowData.read && this.rowData.read.includes('public');

danieltruong: Unless you think something else might use this check

KitArmstrong: Was more just preference, I don't really like cluttering the HTML with a bunch of logic. I don't think it will grow though so I will move it there as it is small.

PR #439, opened 2020-06-19
NRPT-103: Mines Details Page

MaxWardle:


angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.html
```diff
@@ -1,65 +1,96 @@
...
+          <div id="permittee"><b>Permittee: </b>{{mine?.permittee || '-' }}</div>
+        </div>
+        <div class="row label-pair ml-1 mb-2">
+          <div id="status"><b>Operating Status: </b>{{mine?.status || '-'}}</div>
+        </div>
+      </div>
+      <!-- column 2/3 -->
+      <div class="col-lg-4 col-md-12">
+        <div class="row label-pair ml-1 mb-2">
+          <div id="permitNumber"><b>Mines Act Permit Number: </b>{{mine?.permitNumbers && mine?.permitNumbers.length > 0 ? mine?.permitNumbers?.join() : '-'}}</div>
```
dhlevi, line 44: When we have multiple permit numbers, the "join" just stuffs a comma between them. `.join(', ')` will add a space, which is a bit nicer to read.

Second thing: When you have more than a couple, the list can get cluttered. I'm not sure if we're expecting more than a couple normally... if we do it might be nice to restructure the output a bit so it's not just a huge comma separated list. MIght want to ask the UI/UX people about that.

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.html
```diff
@@ -1,65 +1,96 @@
...
-  <div class="row label-pair ml-1 mb-2">
-    <label class="font-weight-bold mr-1">Permittee:</label>
-    <div>{{record?.permittee || '-' }}</div>
+    <label class="font-weight-bold mr-1">Mine Type:</label>
+    <div>{{mine?.type || '-'}}</div>
   </div>
   <div class="row label-pair ml-1 mb-2">
-    <label class="font-weight-bold mr-1">Summary:</label>
-    <div>{{record?.summary || '-' }}</div>
+    <label class="font-weight-bold mr-1">Mine Description:</label>
```
dhlevi, line 75: When mine description/summary are longer, they'll wrap onto a second line. Which is good. It looks strange though when one is wrapped and the other is inline.

Maybe Have the label always display above the text? Just put the text on it's own row and the label above.

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.scss
```diff
@@ -0,0 +1,76 @@
...
+  padding: 9px 26px 14px;
+  box-sizing: content-box;
+
+  @media screen and (max-width: 768px) {
+    font-size: 1.5rem;
+  }
+}
+
+h2 {
+
```
dhlevi, line 15: nit: remove the space

angular/projects/admin-nrpti/src/app/sidebar/sidebar.component.scss
```diff
@@ -27,20 +27,22 @@
...
+.full-nav {
+  padding: 0 0 4px 5px;
+}
+
+.mine-nav {
   background: #00274e;
-  border-radius: 4px 0 0 4px;
-  padding-bottom: 12px;
-  margin: 4px 0 8px;
+  // border-radius: 4px 0 0 4px;
```
dhlevi, line 36: May as well clean out any unused lines

angular/projects/admin-nrpti/src/app/sidebar/sidebar.component.ts
```diff
@@ -37,15 +39,51 @@ export class SidebarComponent implements OnDestroy {
       });
   }
 
+  // SetActiveSidebarItem() {
```
dhlevi, line 42: Nuke if not needed!

angular/projects/admin-nrpti/src/app/sidebar/sidebar.component.ts
```diff
@@ -37,15 +39,51 @@ export class SidebarComponent implements OnDestroy {
...
     urlArray.shift();
-    this.currentMenu = urlArray[0];
-  }
+    const [mainRoute, mainRouteId, currentMenu] = urlArray;
+
+    this.mainRoute = mainRoute;
+    this.mainRouteId = mainRouteId;
+    this.currentMenu = currentMenu && currentMenu.split(';')[0];
+
+    switch (mainRoute) {
```
dhlevi, line 60: Nit: Could probably just be an `if`, rather than a `switch` with only one `case`?

edit: Looks like you could probably merge it with the `if (mainRoute === 'mines') {` on line 72?

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.html
KitArmstrong: Not sure if this helps or not, but on the list page and Core when they have multiple they stack them in a list.

KitArmstrong, line 44: Even though it might not be throwing a linting error, optional chaining isn't available in our version of TS. We are at 3.2.2 and it was introduced in 3.7.

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.ts
```diff
@@ -15,55 +13,115 @@ import { FactoryService } from '../../services/factory.service';
...
+      && this.mine.location.coordinates.length > 0);
+  }
+
+  publish(): void {
+    this.factoryService
+      .publishRecord(this.mine)
+      .pipe(
+        takeUntil(this.ngUnsubscribe),
+        catchError(error => {
+          console.log('Publish error:', error);
```
KitArmstrong, line 66: Shouldn't be console logging on the front end.

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.ts
```diff
@@ -15,55 +13,115 @@ import { FactoryService } from '../../services/factory.service';
...
+      && this.mine.commodities.length > 0
+      && this.mine.location
+      && this.mine.location.coordinates
+      && this.mine.location.coordinates.length > 0);
+  }
+
+  publish(): void {
+    this.factoryService
+      .publishRecord(this.mine)
+      .pipe(
```
KitArmstrong, line 63: I don't think there is any need for `.pipe` here. An error can be handled in subscribe, so there is not need to combine it with `.takeUntil`. I am pretty sure this could be written something like:

`this.factoryService
      .publishRecord(this.Mine)
      .takeUntil(this.ngUnsubscribe)
      . subscribe(
           // Place your main function here,
           // Place your error handler here in the second position
       );`

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.ts
```diff
@@ -15,55 +13,115 @@ import { FactoryService } from '../../services/factory.service';
...
+      )
+      .subscribe(response => {
+        if (!response) {
+          return;
         }
-      }
-    );
+
+        if (response.code === 409) {
+          // object was already published
```
KitArmstrong, line 77: Does the server respond with more information about this? It should... if it does we should probably let hte user know why it failed.

angular/projects/admin-nrpti/src/app/mines/mines-routing.module.ts
```diff
@@ -51,7 +51,7 @@ const routes: Routes = [
               breadcrumb: null
             },
             resolve: {
-              record: MinesResolver
+              mine: MinesResolver
```
KitArmstrong, line 54: Nice update.

angular/projects/common/src/app/models/bcmi/mine.ts
```diff
@@ -23,7 +23,7 @@ export class Mine {
   permittee:            string;
   summary:              string;
   description:          string;
-  links:                string[];
+  links:                [string[]];
```
KitArmstrong, line 26: Links are an array of string arrays? Or are they just an array of strings?

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.html
dhlevi: Yeah, that's kind of what I was thinking as well. Makes the page a bit longer potentially, but improves the clarity of the list

MaxWardle: Yeah would be interesting to hear their thoughts on this. For now I'll add that space between the values, thanks!

MaxWardle: it seems to work for me locally, strange

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.ts
MaxWardle: I'll look into this

angular/projects/admin-nrpti/src/app/mines/mines-detail/mines-detail.component.ts
MaxWardle: I'll try it out, thanks

NickPhura: Not an RXJS expert, so could be wrong:

The pipe catchError is there to catch errors that happen outside of the observable chain. So in this case, if `unPublishRecord()` throws an error for any reason prior to returning an observable. 

I think current RXJS requires pipe for all of its modifier type addons, like takeUntil?

KitArmstrong: @NickPhura is correct. I am not used to that, but anything version 6 or greater uses modifiers in pipes now.

MaxWardle: I left them as is, thanks guys

PR #441, opened 2020-06-22
NRPT-164: Adding clear keywords button to keyword box

dhlevi:
Adding the clear keyword button to the text box for keyword searches in the record search page.

See ticket [NRPT-164](https://bcmines.atlassian.net/browse/NRPT-164)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
+++ b/angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
@@ -173,6 +173,12 @@ export class RecordsListComponent implements OnInit, OnDestroy {
   }
 
   /**
+   * Clears the active keyword search text
+   */
+  clearSearchTerms() {
+    this.keywordSearchWords = '';
+  }
+  /**
    * Updates the url parameters based on the currently set query and table template params, without reloading the page.
    *
    * @memberof RecordsListComponent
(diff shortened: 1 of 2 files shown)


angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
```diff
@@ -172,6 +172,12 @@ export class RecordsListComponent implements OnInit, OnDestroy {
     });
   }
 
+  /**
+   * Clears the active keyword search text
+   */
+  clearSearchTerms() {
+    this.keywordSearchWords = '';
```
NickPhura, line 179: Does the `X` button need to trigger the search when clicked?

dhlevi: Good question. Not mentioned in the ticket, but kinda makes sense. 

Note that currently that clear advanced search does not re-search either.

NickPhura: Oh gotcha.  If the current clear doesn't trigger the search, then seems fine that this wouldn't either. They should at least be consistent with eachother, I think.

NickPhura: Oh gotcha.  If the current clear doesn't trigger the search, then seems fine that this wouldn't either. They should at least be consistent with eachother, I think.

PR #443, opened 2020-06-22
NRPT-169: Memory limit on MongoDB queries

dhlevi:
Some restructuring to the search aggregation to prevent exceeding mongo memory limits.

Adding a `$project` stage after the `$match` stage to remove unecessary query data:

```
$project: {
      _id: 1,
      _flavourRecords: 1,
      read: 1
}
```

Then repopulating with `$lookup` and `$replaceRoot` in the `$facet` stage. If populate is true, those lookups will also merge into the facet stage.

Note that `allowDiskUse` is still set to true as a safetly net for when we do exceed the size limits, but all queries should now run as expected if it is disabled.

See ticket [NRPT-169](https://bcmines.atlassian.net/browse/NRPT-169)


danieltruong:
Noticeably faster! I also disabled allowDiskUse locally just for fun and it works!

PR #445, opened 2020-06-22
NRPT-162: Implemented ellipsis overflow on all table cells.

danieltruong:


danieltruong:
> Looks nice. One thing I noticed. The table looks to be fixed width (1720 or something) so on a smaller screen (for instance, my laptop) we get a horizontal scroll bar and the last couple columns, as well as the paging, is cut out. Might need some min-widths in places of fixed widths perhaps?
> 
> image

Fixed in 741b93f5e1d9b1f64bbec2768a9ae719d1b0b890

PR #446, opened 2020-06-22
NRPT-140: Fixes roles check.

marklise:


angular/projects/common/src/app/constants/auth.ts
```diff
@@ -0,0 +1,6 @@
+export const ROLES = [
```
dhlevi, line 1: These already exist in common/app/utils record-constants.ts. There's a class there called ApplicationRoles that's used in the UI. Can probably remove that one (or use that one) rather than having it in two places.

marklise: Good catch - might be better to move it out from record constants

KitArmstrong: And in api/src/uils/constants/misc.js...

dhlevi: Well, it's nice to have options!

marklise:
Maybe land as separate PR for general cleanup if @KitArmstrong you're in agreement.

KitArmstrong:
Once this one is merged I will create an Roles cleanup PR.

PR #450, opened 2020-06-24
NRPT-171: Changes Edit Record Flavour Order

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-171

Updates the flavour edit area of records that have more than one flavour to match the display order. LNG -> NRCED.


KitArmstrong:
> Flavour edit areas all appear to be in order now. Did we want to do the same for the details pages?

Details page is already in the correct order. This updates edit to match.

PR #452, opened 2020-06-25
FLCN-25: fixed _epicProjectId being set to null

danieltruong:


angular/projects/admin-nrpti/src/app/records/inspections/inspection-add-edit/inspection-add-edit.component.ts
```diff
@@ -308,7 +308,7 @@ export class InspectionAddEditComponent implements OnInit, OnDestroy {
       inspection['_epicProjectId'] = EpicProjectIds.lngCanadaId;
     } else if (inspection['projectName'] === 'Coastal Gaslink') {
       inspection['_epicProjectId'] = EpicProjectIds.coastalGaslinkId;
-    } else {
+    } else if (this.myForm.controls.projectName.dirty) {
```
NickPhura, line 311: Probably need to update the other record types with this change as well, as they all handle projects the same way.

Also, depending on your preference, you could change these blocks to something like:
```
if(projectName.dirty) {
  set projectName
  assign _epicProjectId as needed
}
```
and then you don't need to extra dirty check

danieltruong: e423feb

PR #453, opened 2020-06-25
Update core importer permits

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-208

Updates the Core importer to pull the correct permit number and corresponding permittee. Changed the mine model as there is only a single permit number now. Created a new migration to update the index to reflect the name on the new model.


api/src/integrations/core/datasource.js
```diff
@@ -313,6 +317,60 @@ class CoreDataSource {
...
+
+      // First, any mines with 'X' as their second character are considered exploratory. Remove them.
+      const nonExploratoryPermits = permits.filter(permit => permit.permit_no[1].toLowerCase() !== 'x');
+
+      // Second, mine must not be historical which is indicated by an authorized year of '9999' on the latest amendment.
+      const validPermitNumbers = []
+      for (const permit of nonExploratoryPermits) {
+        // Confirm that the most recent amendment is not historical, which is always the first index.
+        // If 'null' then it is considered valid.
+        if (!permit.permit_amendments[0].authorization_end_date) {
```
dhlevi, line 351: Might be worth a null check if `permit_amendments` might be null or empty? I assume a null/empty permit_amendments set would be valid regardless?

api/src/integrations/core/datasource.js
```diff
@@ -313,6 +317,60 @@ class CoreDataSource {
...
+          // If it is not '9999' it is considered valid.
+          const authDate = new Date(permit.permit_amendments[0].authorization_end_date);
+          if (authDate.getFullYear !== 9999) {
+            validPermitNumbers.push(permit);
+          }
+        }
+      }
+
+      // There should only be a single record. If there is more then we do not want to continue processing.
+      if (validPermitNumbers.length > 1) {
```
dhlevi, line 364: Could probably move this check into the `for` loop to fail fast rather then looping over any other docs.

KitArmstrong: Right, I was dealing with many records before and didn't move this. Moved error into the loop.

api/src/integrations/core/datasource.js
KitArmstrong: To get here if has to have them, but I did add a check just to be safe.

PR #455, opened 2020-06-26
NRPT-227, NRPT-230: Fixes dateAdded and addedBy clobbering on subsequ…

marklise:
…ent importer runs.  Added note and stub about setting a RefId for a mine.  Fixes runs where the API calls post without setting auth_payload correctly.

Couldn't split out the bugs as they are directly related to each other. auth_payload will be set on every request so no need for check-then-default to SYSTEM_USER. Added some documentation on a sample mine record, there might be a better spot for it.

### DIFF
--- a/api/src/integrations/core/datasource.js
+++ b/api/src/integrations/core/datasource.js
@@ -65,7 +65,7 @@ class CoreDataSource {
         _schemaName: "Mine",
         _sourceRefId: "abc123",
         name: "Test Mine",
-        permitNumbers: ['M-209'],
+        permitNumber: 'M-209',
         mine_status: ['Abandoned'],
         mine_type: 'Gold',
         mine_tailings_storage_facilities: ['x'],
--- a/api/src/integrations/core/datasource.js
+++ b/api/src/integrations/core/datasource.js
@@ -26,10 +26,12 @@ class CoreDataSource {
    * Creates an instance of CoreDataSource.
    *
    * @param {*} taskAuditRecord audit record hook for this import instance
+   * @param {*} auth_payload information about the user account that started this update.
    * @memberof CoreDataSource
    */
-  constructor(taskAuditRecord) {
+  constructor(taskAuditRecord, auth_payload) {
     this.taskAuditRecord = taskAuditRecord;
+    this.auth_payload = auth_payload;
 
     // Set initial status
     this.status = { itemsProcessed: 0, itemTotal: 0, individualRecordStatus: [] };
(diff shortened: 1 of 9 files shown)


api/src/integrations/core/datasource.js
```diff
@@ -25,10 +25,12 @@ class CoreDataSource {
    * Creates an instance of CoreDataSource.
    *
    * @param {*} taskAuditRecord audit record hook for this import instance
+   * @param {*} auth_payload information about the user account that started this update.
    * @memberof CoreDataSource
    */
-  constructor(taskAuditRecord) {
+  constructor(taskAuditRecord, auth_payload) {
```
KitArmstrong, line 31: On the other objects the auth_payload is the param. Might be nice to make them align.

api/src/integrations/core/mine-utils.js
```diff
@@ -58,6 +59,8 @@ class Mines extends BaseRecordUtils {
       location : { type: 'Point', coordinates: mineRecord.coordinates },
       permittee: this.getParty(MINE_PARTY_PERMITTEE, mineRecord, permit),
       type: '',
+      // TODO:  Without this, subsequent runs of the mine importer will duplicate mines every time it's run
```
KitArmstrong, line 62: Does this still need to be done?

api/src/integrations/core/datasource.js
```diff
@@ -58,6 +60,20 @@ class CoreDataSource {
 
   /**
    * Main function that runs all necessary operations to update Core records.
+   * Sample Record Data for reference
+      {
+        _schemaName: "Mine",
+        _sourceRefId: "abc123",
+        name: "Test Mine",
+        permitNumbers: ['M-209'],
```
KitArmstrong, line 68: it is now `permitNumber: string`.

api/src/integrations/core/mine-utils.js
marklise: Yes

api/src/integrations/core/datasource.js
marklise: I wanted to limit the amount of adjustments I made - cleanup ticket perhaps?

api/src/integrations/core/datasource.js
marklise: Fixed in e40a974

api/src/integrations/core/mine-utils.js
marklise: Ticket: https://bcmines.atlassian.net/browse/NRPT-232

PR #459, opened 2020-06-29
NRPT-233: Updates Core importer for better error handling

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-233

I made multiple changes to better import the Core records and better handle errors.

- New parameter to only import major mines (cuts down from 18,000 to ~110 records)
- Makes subsequent API calls to get more details about each record to only send 2 at a time (fixes 504 errors)
- If there are any API failures or if a record doesn't contain valid data the Mine ID is logged but the other records continue to process

When running against prod data there are 14 records that will throw errors. These are data related and are to be expected right now.

### DIFF
--- a/api/src/integrations/core/datasource.js
+++ b/api/src/integrations/core/datasource.js
@@ -114,60 +114,10 @@ class CoreDataSource {
    */
   async getAllRecordData() {
     try{
-      let currentPage = 1;
-      let totalPages = 1;
-      let mineRecords = [];
+      const verifiedMines = await this.getVerifiedMines();
+      const minesWithDetails = await this.addMinesDetails(verifiedMines);
 
-      // The Core API can not return all data in a single call. Must fetch data in batches.
-      do {
-        const queryParams = { per_page: CORE_API_BATCH_SIZE, page: currentPage };
-        const url = this.getIntegrationUrl(CORE_API_HOST, CORE_API_PATH_MINES, queryParams);
-  
-        // Get Core records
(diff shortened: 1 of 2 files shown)


NickPhura:
Totally optional (apparently I can't comment inline, as its not part of the PR):

Since `getAllRecordData()` is chaining the calls to build up the complete array of coreRecords, you could also move the call to get commodities to its own function and call it in a similar manner to the `getVerifiedMines` and `addMinesDetails` calls.  Not sure how much code would need to be changed to make that work though, so potentially not worth it.

api/src/integrations/core/datasource.js
```diff
@@ -392,6 +347,91 @@ class CoreDataSource {
...
+      const mineDetailsUrl = this.getIntegrationUrl(CORE_API_HOST, mineDetailsPath);
+
+      try {
+        const [ parties, mineDetails ] = await Promise.all([
+          integrationUtils.getRecords(partyUrl, this.getAuthHeader()),
+          integrationUtils.getRecords(mineDetailsUrl, this.getAuthHeader())
+        ]);
+
+        const latitude = mineDetails.mine_location && mineDetails.mine_location.latitude || 0.00;
+        const longitude = mineDetails.mine_location && mineDetails.mine_location.longitude || 0.00;
```
NickPhura, line 421: Does `mineDetails` need a null check?

KitArmstrong: Added.

PR #460, opened 2020-06-29
NRPT-188: OGC Inspection CSV importer + Tests

NickPhura:
Ticket: https://bcmines.atlassian.net/browse/NRPT-188

There is a CSV attached, but also, you  can generate a new one (with the latest data) by going to this URL.  This is ultimately what the admin users will do when importing new data:
https://reports.bcogc.ca/ogc/f?p=200:501::CSV


api/src/importers/bcogc/base-record-utils.js
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Searches for an existing master record, and returns it if found.
+   *
+   * @param {*} nrptiRecord
+   * @returns {object} existing NRPTI master record, or null if none found or _sourceRefOgcInspectionId &&
+   * _sourceRefOgcDeficiencyId are null
+   * @memberof BaseRecordUtils
+   */
+  async findExistingRecord(nrptiRecord) {
+    if (!nrptiRecord._sourceRefOgcInspectionId && !nrptiRecord._sourceRefOgcDeficiencyId) {
```
KitArmstrong, line 65: Just confirming that this is indeed an "and" rather than "or"?

NickPhura: Actually, now that I think about it more, the DeficiencyId null check should be removed.

The way to uniquely identify the records is from either:
- InspectionId=`uniqueString` && DeficiencyId=`null`
- InspecitonId=`nonUniqueString` && DeficiencyId=`uniqeStringForThisInspectionId`

So we really only care that InspectionId is not null.

NickPhura: Though, its an extreme edge case, as the only issue with the current setup is if InspectionId is null and DeficiencyId is not null, but that shouldn't ever be the case.  Ill update it anyways though since it is technically safer.

PR #462, opened 2020-06-29
NRPT-156: Add collection model, change mine -> minebcmi

marklise:


api/src/models/bcmi/collection.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+module.exports = require('../../utils/model-schema-generator')(
+  'BCMICollection',
```
dhlevi, line 4: Is there any desire/future potential for collections to be needed by any other system? I doubt it, but if so maybe DocCollection or something rather than specifying BCMI?

Also, really picky, but we use MineBCMI for the Mines, maybe this should be CollectionBCMI just for consistency in the naming style?

marklise: Snap - good catch.

marklise: Fixed in 7511fa3

PR #467, opened 2020-07-02
NRPT-117: Adds Permit and Amendment import with flavours

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-117

This updates the Core importer to create Permits and PermitAmendments for all Mine records.  Mines are only created as a BCMI flavour (previous change, not this ticket), but the Permits and PermitAmendments have both master and BCMI flavours. This is because it might be required to share this data in the future, especially the permit as it does exist for LNG. 

If running the importer against the Core Prod API (recommended) there will be about 15 records that it errors on. This is expected and should be logged to the console.


api/src/models/master/permitAmendment.js
```diff
@@ -0,0 +1,33 @@
...
+    typeCode: { type: String, default: '' },
+    receivedDate: { type: Date, default: null },
+    issueDate: { type: Date, default: null },
+    authorizedEndDate: { type: Date, default: null },
+    description: { type: String, default: '' },
+    documents: [{ type: 'ObjectId', default: [], index: true }],
+
+    dateAdded: { type: Date, default: Date.now() },
+    dateUpdated: { type: Date, default: null },
+    datePublished: { type: Date, default: null },
```
NickPhura, line 24: Can remove the `datePublished` and `publishedBy` from the master record, as they should never themselves get published.

api/src/models/master/permitAmendment.js
```diff
@@ -0,0 +1,33 @@
...
+
+    addedBy: { type: String, default: '' },
+    updatedBy: { type: String, default: '' },
+    publishedBy: { type: String, default: '' },
```
NickPhura, line 28: Could add `sourceDateAdded` and `sourceDateUpdated` to be consistent with the other types.

Also, probably need to add `isBcmiPublished` ?  This would be in alignment with the other types, though I'm not sure how collections influences this, if it does.

api/src/models/master/permit.js
```diff
@@ -31,6 +31,11 @@ module.exports = require('../../utils/model-schema-generator')(
     centroid: [{ type: Number, default: 0.0 }],
     documents: [{ type: 'ObjectId', default: [], index: true }],
 
+    mineGuid: { type: String, default: '' },
+    permitNumber: { type: String, default: '' },
+    status: { type: String, default: '' },
+    permitAmendments: [{ type: 'ObjectId', default: [], index: true }],
+
```
NickPhura, line 38: Similar to my below comment, could add `isBcmiPublished` to align with the other models, though not sure how collections impacts this.

api/src/models/bcmi/permit-bcmi.js
```diff
@@ -0,0 +1,26 @@
+module.exports = require('../../utils/model-schema-generator')(
+  'PermitBCMI',
+  {
+    _schemaName: { type: String, default: 'PermitBCMI', index: true },
+    _sourceRefId:  { type: String, default: null, index: true },
+
+    read: [{ type: String, trim: true, default: 'sysadmin' }],
+    write: [{ type: String, trim: true, default: 'sysadmin' }],
+
+    mineGuid: { type: String, default: '' },
```
NickPhura, line 10: Open question: do we need to include all of the other `Permit` fields here as well? Our other records always copy all of the master fields, regardless of whether or not they get shown on the frontends.  But maybe that isn't necessary? Not sure

api/src/integrations/core/permit-amendment-utils.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    return  permit.permit_amendments.map(amendment => ({
+      ...super.transformRecord(permit),
+      _sourceRefId: amendment.permit_amendment_guid || '',
+      statusCode: amendment.permit_amendment_status_code || '',
+      typeCode: amendment.permit_amendment_type_code || '',
+      receivedDate: (amendment.received_date && new Date(amendment.received_date)) || null,
+      issueDate: (amendment.issue_date && new Date(amendment.issue_date)) || null,
+      authorizedEndDate: (amendment.authorized_end_date && new Date(amendment.authorized_end_date)) || null,
```
NickPhura, line 50: Probably a nit: Assuming the dateString is in a valid form, we could probably just set it directly without calling new Date()?

api/src/integrations/core/datasource.js
```diff
@@ -295,57 +309,132 @@ class CoreDataSource {
...
+
+    if (!permit) {
+      throw new Error('updateMinePermit - Cannot find valid permit');
     }
+
+    // Update the amendments and permit.
+    const transformedAmendments = await permitAmendmentUtils.transformRecords(permit);
+    // To trigger flavour for this import.
+    const preparedAmendments = transformedAmendments.map(amendment => ({ ...amendment, PermitAmendmentBCMI: {} }))
+    const updatedAmendments = await permitAmendmentUtils.updateRecord(preparedAmendments);
```
NickPhura, line 425: Is it possible to be performing an update to an existing Permit, but have new amendments that need to get created?
If so, we probably need to do the findExisting check here first, and the call either permitAmendmentUtils update/create.

api/src/controllers/put/mine-permit.js
```diff
@@ -0,0 +1,74 @@
+const mongoose = require('mongoose');
```
NickPhura, line 1: Do we still have the concept of a `mine-permit`, or is this replaced by `permit` and `permt-amendment`?

KitArmstrong: Do not. Removed.

api/src/integrations/core/permit-amendment-utils.js
KitArmstrong: Updating.

api/src/models/bcmi/permit-bcmi.js
KitArmstrong: I was talking to @marklise about this and opted to follow what the others have done. However, I would think they should just live on the flavour.

NickPhura: If I understand what you said correctly (possibly I don't), then we DO want to add all of the other permit fields here? `recordName`, `dateIssued`, `legislation`, etc.

api/src/models/master/permitAmendment.js
KitArmstrong: Added `isBcmiPublished` to permit and permit amendments. I don't capture the source date information as none of that is returned by the Core API.

api/src/models/master/permitAmendment.js
KitArmstrong: Removed from master.

api/src/models/master/permit.js
KitArmstrong: Added.

api/src/integrations/core/datasource.js
KitArmstrong: Good call. Updated to handle an update and create.

api/src/models/bcmi/permit-bcmi.js
KitArmstrong: Oh sorry, I thought this was on the master. Umm, to me it seems unnecessary on the flavour, but maybe there was a reason it was done before. Do you think they should be? Would they be used for anything?

NickPhura: > Do you think they should be? Would they be used for anything?

Not sure.  Trying to think if there is a good reason to include them, other than just because that is what was done in past. 

The only thing that comes to mind, is if we wanted to use those other values for filtering/searching. 
On LNG, we don't show that many fields, but some of them get used for searching/filtering.  BCMI in theory might be similar, though with collections, its harder to say, since now we aren't directly querying for bcmi records from the public site.

So not sure :/

I guess given that we have no current reason to change it, we can just leave it as is, and always migrate it over in future if it becomes necessary for w/e reason

api/src/models/master/permitAmendment.js
NickPhura: > I don't capture the source date information as none of that is returned by the Core API.

Yeah, fair enough then

api/src/models/bcmi/permit-bcmi.js
KitArmstrong: Okay, that sounds good.

KitArmstrong:
@NickCorcoran Addressed your comments.

NickPhura:
> 
> 
> @NickPhura Addressed your comments.

Sweet, will give it another look over in a bit here

NickPhura:
@KitArmstrong I tried running the core importer, but got a 503 error.  It looks like I did successfully got an api token (says it expires in 16 hours), but then the subsequent calls to get verified mines, etc, returned 503.

I've got all of the env vars set up for Core Test.  Maybe it is down? Or is there possibly some other config I am missing?

Ill mess with it a bit more though and see what I come up with

NickPhura:
I know this is a known side effect of Core not supplying any of the 'normal' Permit values, but we'll probably want to put some kind of canned data into some of these fields, so it doesn't render empty like this (see screenshot).

Probably this is something that UX/PO folks need to think about, so maybe can be its own ticket.

image

api/src/models/master/permitAmendment.js
marklise: If you add it here, `isBcmiPublished` needs to be done for all other master records in order to facilitate search on the admin page.

PR #471, opened 2020-07-02
NRPT-62: Remove fullRecord, redact after doc/flavour import

dhlevi:
It was possible to float documents/flavours out of the API when a user lacked the required roles due to the redaction occurring before the population of documents. An additional redaction step has been added to the document and flavour populate to re-redact these if the user does not have the roles.

I've left the original redaction pipe in place where it is, just to preemptively strip out records we don't need to process, and keep behaviour consistent when populate=false.

See ticket [NRPT-62](https://bcmines.atlassian.net/browse/NRPT-62)

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -329,7 +329,43 @@ let searchCollection = async function (
         as: 'documents'
       }
     });
+
+    // Redaction. We've imported details from
+    // flavours and documents, and we may need
+    // to prevent some of these from being returned
+    // if the user lacks the requisite role(s)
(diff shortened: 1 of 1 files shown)


NickPhura:
Probably need to add that second redaction step to the get item search as well, as it also populates documents after the original redaction. Around line 600ish.

dhlevi:
> Probably need to add that second redaction step to the get item search as well, as it also populates documents after the original redaction. Around line 600ish.

Yeah good catch. Those ones are all inline so I can just move the redact after the populate. It'll wind up identical if populate=false.

PR #476, opened 2020-07-03
NRPT-207: Update Tailings Storage Facilities when blank in Core

marklise:


marklise:
> Just a thought... this will change null values to N/A, but not zero. Was the intention to catch zero as well?

The way I read the ticket was for non-zero, non-integer values = N/A.  It's a good question - I'll pose the question on JIRA: https://bcmines.atlassian.net/browse/NRPT-207

PR #479, opened 2020-07-06
NRPT-220: Update so that all admin rolls can use the importers, and see import history

MaxWardle:
https://bcmines.atlassian.net/browse/NRPT-220


api/src/controllers/document-controller.js
```diff
@@ -254,15 +254,15 @@ exports.protectedDelete = async function(args, res, next) {
...
+async function createURLDocument(fileName, addedBy, url, readRoles = ['admin:lng', 'admin:nrced', 'admin:bcmi']) {
   const Document = mongoose.model('Document');
   let document = new Document();
 
   document.fileName = fileName;
   document.addedBy = addedBy;
   document.url = url;
   document.read = ['sysadmin', ...readRoles];
-  document.write = ['sysadmin'];
+  document.write = ['sysadmin', 'admin:lng', 'admin:nrced', 'admin:bcmi'];
```
KitArmstrong, line 265: I am not sure that this is the approach we want to take. For instance, when I am adding new documents for BCMI I do not want other admins to have write permission on them. I think it makes sense for the task stuff that is below, just not sure about documents.

MaxWardle: Yes I agree, however the last line of the [ticket](https://bcmines.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=NRPT-220) is "all roles can import and see import history list". which isn't possible without them having write permissions on docs. The importer will run but none of the docs will update.

MaxWardle: Maybe more conversation needed here

NickPhura: Some more context:

The main issue, that needs to be addressed, whether as above or in some other way, is that documents are always auto published/unpublished whenever a record is created/updated.  

Specifically, a document is published if the record does not contain information about an underage person, and is never published if it does.  So regardless of who creates the record, the document gets published/unpublished every time a record is created/updated (since whenever the data changes, we need to re-assess if it contains sensitive data or not).

So regardless of whether it is a record coming from an importer, or a record manually added, anyone that has access to that record needs to have write access to the document(s).  Because as soon as they cause a change in the records data, the documents will be published/unpublished, and if the users write role is missing, it will fail to do so.

So I think ultimately, documents need to get all of the roles in the same way that records get all of the roles (assuming they currently do get all of the roles, I think they do?).

api/src/controllers/document-controller.js
```diff
@@ -254,15 +254,15 @@ exports.protectedDelete = async function(args, res, next) {
  * @param {*} [url=null] url of the document
  * @param {*} [readRoles=[]] read roles to add to the document. (optional)
  */
-async function createURLDocument(fileName, addedBy, url, readRoles = []) {
+async function createURLDocument(fileName, addedBy, url, readRoles = ['admin:lng', 'admin:nrced', 'admin:bcmi']) {
```
NickPhura, line 257: Currently, the readRoles array only is used to set `public`, so these defaults will always get overridden.

I think we would just want to set those 3 roles directly in the document every time, in the same way the write roles are set.

NickPhura: `  document.read = ['sysadmin', 'admin:lng', 'admin:nrced', 'admin:bcmi', ...readRoles];`

NickPhura: Assuming that my blub about documents needing all roles is accurate, and i overlooked anything.

api/src/controllers/document-controller.js
KitArmstrong: Looks like it will be okay for all admins to have write.

marklise: Documents are not (and should not be) auto-published across the board.  Documents coming in from CORE will have to remain non-public as they may contain PIA and will require a human to either approve for publishing, or redact and create a new version for use.  So this is an incoming change.  The rule here is that any of those admin:*/sysadmin roles have the ability to make that happen as they *should* be consistent between the business areas.

I don't believe that we should be setting the roles default in the constructor, this should be always empty and leveraged appropriately.  In the setting where you do ```document.write = ['sysadmin', 'admin:lng', 'admin:nrced', 'admin:bcmi'];``` I think we could use a constant declared in `misc.js`.  See ROLES.ADMIN_ROLES export:

```
exports.ROLES = {
  SYSADMIN: 'sysadmin',
  LNGADMIN: 'admin:lng',
  NRCEDADMIN: 'admin:nrced',
  BCMIADMIN: 'admin:bcmi',

  ADMIN_ROLES: [
    'sysadmin',
    'admin:lng',
    'admin:nrced',
    'admin:bcmi'
  ]
};
```

NickPhura: > Documents are not (and should not be) auto-published across the board. Documents coming in from CORE will have to remain non-public as they may contain PIA and will require a human to either approve for publishing, or redact and create a new version for use. So this is an incoming change. 

Yeah, this will require a bunch of changes to support. Is this already something in a future ticket?

Currently, documents are auto-published base on the business logic anonymous rules (business-logic-manager.js).  And we don't have any way for a user to manually publish/unpublish a document.

NickPhura: So we could add an exception to `business-logic-manager -> isDocumentConsideredAnonymous` to make documents, whose record has sourceSystemRef === 'core' , anonymous.  But we would still need some mechanism to undue that later. Possibly when a user edits a record, if the source-system changes to NRPTI, then it would pass, so maybe that would work as is?

api/src/controllers/document-controller.js
marklise: As for the code change above, no default constructor variables needed - so remove the auto-assignment.  Leave `readRoles = []` and add `writeRoles =[]`.

As for the comment above, we shouldn't hard code roles inline like that.  Let the caller of the function define what they should be so there's less automatic behaviors we might need to unwind later if business rules change.

```document.read = ['sysadmin', ...readRoles];```
and
```document.write= ['sysadmin', ...writeRoles];```

File: api/src/controllers/document-controller.js [hunk 3]
```diff
@@ -261,8 +264,8 @@ async function createURLDocument(fileName, addedBy, url, readRoles = []) {
   document.fileName = fileName;
   document.addedBy = addedBy;
   document.url = url;
-  document.read = ['sysadmin', ...readRoles];
-  document.write = ['sysadmin'];
+  document.read = [ROLES.SYSADMIN, ROLES.LNGADMIN, ROLES.NRCEDADMIN, ROLES.BCMIADMIN, ...readRoles];
+  document.write = ROLES.ADMIN_ROLES;
```
KitArmstrong, line 268: This should probably match the function below.

MaxWardle: nice catch, pushed up a change

api/src/controllers/document-controller.js
```diff
@@ -257,15 +258,15 @@ exports.protectedDelete = async function(args, res, next) {
...
   const Document = mongoose.model('Document');
   let document = new Document();
 
   document.fileName = fileName;
   document.addedBy = addedBy;
   document.url = url;
-  document.read = [ROLES.SYSADMIN, ROLES.LNGADMIN, ROLES.NRCEDADMIN, ROLES.BCMIADMIN, ...readRoles];
-  document.write = ROLES.ADMIN_ROLES;
+  document.read = [ROLES.SYSADMIN, ...readRoles];
+  document.write = [ROLES.SYSADMIN, ...writeRoles];;
```
marklise, line 269: `const` here

PR #483, opened 2020-07-08
NRPT-248: Add a published field to search results

dhlevi:
Adding a published field to search results to support sorting by published.

See ticket [NRPT-248](https://bcmines.atlassian.net/browse/NRPT-248)

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -263,8 +263,8 @@ let searchCollection = async function (
               }
             ]
           },
-          then: true,
-          else: false }
+          then: 'published',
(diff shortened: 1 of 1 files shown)


dhlevi:
> Works. NB: given the screenshot in NRPT-248, the 1+published1 sort will have `unpublished` sorted first, and the `published` items at the end. For the screenshot to be accurate, either the triangle should be pointing down, or the `published` results not showing (ie: in the bottom of the list).

Yeah good catch. Sort is string based so false comes before true. I'll change the results of the published step to be published/unpublished instead of true/false

PR #486, opened 2020-07-13
NRPT-199: Common Search Component

dhlevi:
Creation of a common search component for use with NRPTI admin.

The common component replaces existing keyword search and explore page functionality and screens. A readme has been created to assist with implementing the component elsewhere, but in case nobody wants to read it, classes and components are fairly well documented (see documentation [readme here](https://github.com/dhlevi/NRPTI/blob/NRPT-199-search-comp/angular/projects/common/src/app/search-filter-template/readme.md)).

The component can be found in the common project under `search-filter-template`. Fairly straightforward to use:

```html
  <search-filter-template
    (searchEvent)="search($event)"
    (resetControls)="filterReset($event)"
    (filterChange)="filterChange($event)"
    [title]="myTitleVar"
    [tooltip]="myooltipVar"
    [subsets]="subsets"
    [advancedFilters]="true"
    [advancedFilterTitle]="myAdvancedFilterTitle"
    [advancedFilterText]="myFilterDesc"
    [attachPanelToDiv]="myHostComponentDiv"
    [showAdvancedFilters]="false"
    [filters]="filters"
    [searchOnFilterChange]="true">
  </search-filter-template>
```

Include the above to add a search component to a screen (some of the attributes are optional of course, check the docs). To set up subset searches just add a `SubsetObject`:

```typescript
const subsetOptions = [
  new SubsetOption('', 'All'),
  new SubsetOption('issuedTo', 'Issued To'),
  new SubsetOption('location', 'Location'),
  new SubsetOption('description', 'Description & Summary'),
  new SubsetOption('recordName', 'Record Name')
];
this.subsets = new SubsetsObject(subsetOptions);
```

and to setup filters, just add a `FilterObject` with the approriate type and `FilterDefinition` in place. You can add one or more for advanced filters.

```typescript
const entityTypeFilter = new FilterObject(
  'entityType',
  FilterType.Checkbox,
  'Entity Type',
  new CheckOrRadioFilterDefinition([new OptionItem('issuedToCompany', 'Company'), new OptionItem('issuedToIndividual', 'Individual')])
);
```

Unit test is included. To "prove" the component works and covers the bases for our existing set of filters, I've also replaced the existing explore page and search component on the records-list component and the mines-list components, so you can see the search template in action by checking those pages out. All filter types required by the Mines components are included, as well as others needed for the records page.

Some obvious points of extension are some additional filter types, and allowing for custom CSS on the components. Extending filter definitions should be pretty easy as the need arises.

Including everyone for review as it's a big component and a a lot of stuff to look over and I undoubtedly missed something or left in a bug or two!

See Ticket [NRPT-199](https://bcmines.atlassian.net/browse/NRPT-199)


angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.html
```diff
@@ -1,39 +1,13 @@
...
-            </button>
-          </div>
-        </div>
-      </div>
-    </div>
-  </div>
+  <search-filter-template
+    #explorePanel
+    (searchEvent)='executeSearch($event)'
+    [title]="'Keyword Search'"
```
KitArmstrong, line 7: Just confirming you intended the inner single quote in these strings.

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
```diff
@@ -103,7 +105,175 @@ export class MinesRecordsListComponent implements OnInit, OnDestroy {
...
+      responsibleAgencyFilter,
+      issuedUnderRegFilter,
+      sourceSystemFilter,
+      projectFilter,
+      documentsfilter
+    ];
+  }
+
+  executeSearch(searchPackage) {
+    console.log(searchPackage);
```
KitArmstrong, line 234: Remove.

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
```diff
@@ -103,7 +105,175 @@ export class MinesRecordsListComponent implements OnInit, OnDestroy {
...
+    Object.keys(searchPackage.filters).forEach(filter => {
+      this.queryParams[filter] = searchPackage.filters[filter];
+    });
+
+    this.tableData.currentPage = 1;
+    this.submit();
+  }
+
+  private clearQueryParamsFilters() {
+    delete this.queryParams['keywords'];
```
KitArmstrong, line 260: Are there other params that need to remain or can `this.queryParams` just be set to an empty object?

angular/projects/common/src/app/search-filter-template/filter-object.ts
```diff
@@ -0,0 +1,212 @@
+import { IMutliSelectOption } from '../autocomplete-multi-select/autocomplete-multi-select.component';
+
+/**
```
KitArmstrong, line 3: Nice documentation on this. Was easy to follow how to use it.

angular/projects/common/src/app/search-filter-template/search-filter-template.component.ts
```diff
@@ -0,0 +1,337 @@
...
+    // ensure we parse through values from the URL and preselect anything
+    // that needs pre-selecting
+
+    if (this.router && this.router.url) {
+      this.router.url.split(';').forEach(filterVal => {
+        if (filterVal.split('=').length === 2) {
+          const filterName = filterVal.split('=')[0];
+          const val = filterVal.split('=')[1];
+
+          if (val && val !== 'null' && val.length !== 0) {
```
KitArmstrong, line 83: `if(val)` accomplishes the same thing here.

angular/projects/admin-nrpti/src/app/mines/mines-list/mines-list.component.html
dhlevi: Removed the square brackets and the single quotes. Didn't need to parse for a var in these values on the screen

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
dhlevi: Removed

angular/projects/common/src/app/search-filter-template/search-filter-template.component.ts
dhlevi: fixed

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
dhlevi: Yeah, looks like in the current screens the queryParams holds some other values related to paging, etc. Originally I had it set to just do a `this.queryParams = {}` but that was causing issues, so this is a safety net for that. A better fix is to probably ensure queryParams is just for query params, and tableParams is handling all the table params without any crossover; checking into them now.

dhlevi:
Note: Added a couple new features that may come in handy in the future:

- You can now set the default size for each advanced filter component at the 'root' level, using the col class values of 1-12 (defaults to 4 to match records list page). These are also overridable at the FilterObject level, so you can set a default of, say 6 for all components, and set a specific one to 12 to cover the whole row. 
- Added event emitter when the advanced filter panel is toggled.
- Added title and subtext to the advanced filter panel
- Added the ability to append advanced filter panel to another div if you want to handle it on the host component more directly

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.html
```diff
@@ -1,130 +1,15 @@
...
-      <app-explore-panel
-        *ngIf="searchFiltersForm"
-        #explorePanel
-        [formGroup]="searchFiltersForm"
-        (filtersReset)="clearKeywordSearch(); resetSortBy()"
-      ></app-explore-panel>
-    </div>
-  </div>
-</section>
+  <search-filter-template
```
MaxWardle, line 2: woowee thats nice

NickPhura:
When you hover over any of the filter elements, or the `Open Advanced Filters` button, the `title` popup always says 'Keyword Search'

NickPhura:
Minor weirdness in the radio button alignment

image

NickPhura:
Did some testing for the filters when ALL of them are set, etc:

- The `Reset Filters` works almost fully.  The only thing that doesn't get reset is the UI for the date range pickers.  The URL is cleared correctly, the date pickers just still show the date string in the UI (ie: '202-05-03').

- Some filters aren't set in the UI, if you copy the url into a new tab.
  - For example: if the responsible agency filter is set and the list gets filtered.  Copying the url into a new tab results in the list being filtered correctly, but the UI control doesn't visually indicate that anything is being filtered.
  - This seems to be the case for the `subset`, `act`, `regulation`, and `responsible agency` filters. The others seem to work correctly.

dhlevi:
> Radio button labels are appearing on a new line for some reason.

Interesting. For me they are in test as well, so I didn't notice anything different. I'll see if I can fiddle with the spacing and ensure they're on the same line, but it does appear to be consistent with the original.

NickPhura:
> 
> 
> > Radio button labels are appearing on a new line for some reason.
> 
> Interesting. For me they are in test as well, so I didn't notice anything different. I'll see if I can fiddle with the spacing and ensure they're on the same line, but it does appear to be consistent with the original.

In Test, using Firefox, they appear on the same line initially, and then as the window gets smaller they drop down. In this branch they are just always dropped-down.  Its a pretty minor difference though, as even in test, they start to drop down at around 1740, which isn't much less than 1920.

dhlevi:
> * The `Reset Filters` works almost fully.  The only thing that doesn't get reset is the UI for the date range pickers.  The URL is cleared correctly, the date pickers just still show the date string in the UI (ie: '202-05-03').

Looks like the date picker control was not handling a form reset. Updated that control to listen for resets
```
this.control.statusChanges.subscribe(x => {
      if (this.control.pristine) {
        this.clearDate();
      }
    });
```

> 
> Testing when you copy the URL into a new tab:
> 
> * Some filters aren't set in the UI, if you copy the url into a new tab.

Looks like a bug in the form control rebuilder. It's mapping displayValue/Value wrong in some cases on the `IMutliSelectOption`. Fixing now!

dhlevi:
> Looks like a bug in the form control rebuilder. It's mapping displayValue/Value wrong in some cases on the `IMutliSelectOption`. Fixing now!

Just put up a fix for this. Issue was encoded values not mapping back to their unencoded counterpart for display. Added a decodeUriComponent call to dropdown/multiselect strings to fix this.

dhlevi:
> Radio button labels are appearing on a new line for some reason.

Upped the column size from 3 to 5. Should correct the issue. Originally I tried just getting rid of column spacing altogether and do it manually, but that could lead to situations where one column was dropping below before the other and it looked a bit off.

dhlevi:
> Regarding my previous comments: the only thing left is the subset filter UI. It doesn't get set on url copy/paste.

Fixed!

PR #492, opened 2020-07-16
NRPT-212: Adds functionality to front-ends, supports  on the back-end…

marklise:
…, and adds the mine -> epic project mapping migration/lookup.

NB: This can't be merged or run in test/prod until the `permitNumber: ""` issue is fixed: https://bcmines.atlassian.net/browse/NRPT-252


angular/projects/global/src/lib/services/search.service.ts
```diff
@@ -100,6 +101,11 @@ export class SearchService {
         });
       });
     }
+    if (_in && _in !== {}) {
```
KitArmstrong, line 104: This will always pass because `_in` will always not equal `{}`.  `{} === {}` will always be false.  If checking for any empty object this way it will need to be converted to a string first or another way will need to be used.

`JSON.stringify(_in) !== '{}'`  or
`Object.keys(_in).length !== 0`

marklise: Good catch! Fixed in d2bcc9b

danieltruong:
> Cannot compile Admin, I get the following error:
> `ERROR in projects/admin-nrpti/src/app/services/factory.service.ts(256,12): error TS2554: Expected 4-12 arguments, but got 13.`

You might have to npm run build:library:dev for that one.

KitArmstrong:
> > Cannot compile Admin, I get the following error:
> > `ERROR in projects/admin-nrpti/src/app/services/factory.service.ts(256,12): error TS2554: Expected 4-12 arguments, but got 13.`
> 
> You might have to npm run build:library:dev for that one.

That fixed it. Thanks.

PR #501, opened 2020-07-20
NRPT-196: Mine Collections add-edit

NickPhura:
Ticket: https://bcmines.atlassian.net/browse/NRPT-196

- Add mine collections add-edit component
- Update search.js to populate collection records
- Add collection post/put controllers
- Misc tweaks

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
@@ -1,0 +1,251 @@
+<main class="container-fluid-padding">
+  <form *ngIf="!loading" [formGroup]="myForm" novalidate>
+    <section>
+      <div class="d-flex heading-bar">
+        <div>
+          <h1 id="name" class="m-0">
+            {{ (isEditing && collection && collection.name) || (!isEditing && myForm.get('collectionName').value) }}
+          </h1>
+        </div>
+      </div>
+    </section>
+
+    <section class="mb-3 pb-3 border-bottom">
+      <div class="d-flex justify-content-between">
+        <div class="col-6 col-lg-8 col-xl-9">
+          <div>
+            <h2 class="border-0 mb-0">Edit Collection</h2>
+          </div>
+          <div class="mb-3">
+            <span class="grey-subtext">{{ lastEditedSubText }}</span>
+          </div>
+          <div>
+            <mat-slide-toggle
+              name="collectionPublish"
+              id="collectionPublish"
+              class="publish-toggle"
+              [ngClass]="myForm.get('collectionPublish').value ? 'published' : 'unpublished'"
+              formControlName="collectionPublish"
+              (change)="togglePublish($event)"
+              [checked]="myForm.get('collectionPublish').value"
+              title="{{ myForm.get('collectionPublish').value ? 'Unpublish collection' : 'Publish collection' }}"
+              ngDefaultControl
+            >
+              {{ (myForm.get('collectionPublish').value && 'Published') || 'Unpublished' }}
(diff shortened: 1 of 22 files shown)


api/src/controllers/post/collection-bcmi.js
```diff
@@ -0,0 +1,84 @@
+const mongoose = require('mongoose');
```
dhlevi, line 1: Do we need any explicit checks for a user having the `admin:bmi` role for these? Looks like from swagger it'll let through `admin:nrced` and `admin:lng` even though those admins shouldn't have access to post a collection?

api/src/controllers/put/collection-bcmi.js
```diff
@@ -0,0 +1,95 @@
+const mongoose = require('mongoose');
```
dhlevi, line 1: Same as the comment for post/collection-bcmi.... role check for admin:bcmi explicity?

angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
```diff
@@ -0,0 +1,251 @@
...
+              ngDefaultControl
+            >
+              {{ (myForm.get('collectionPublish').value && 'Published') || 'Unpublished' }}
+            </mat-slide-toggle>
+          </div>
+        </div>
+        <div class="col-6 col-lg-4 col-xl-3">
+          <div class="d-flex justify-content-end mb-1">
+            <button type="button" class="btn btn-link" (click)="cancel()" title="Cancel">Cancel</button>
+            <button class="btn btn-primary" type="submit" (click)="submit()" title="Update">
```
dhlevi, line 41: Button should say "save" when creating a new record?

api/src/controllers/search.js
```diff
@@ -649,6 +649,20 @@ const executeQuery = async function (args, res, next) {
       }
     ];
 
+    // Populate bcmi collection records
```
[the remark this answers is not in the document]
NickPhura: Yeah I can throw that in there

angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
NickPhura: Yup, will fix

api/src/controllers/post/collection-bcmi.js
NickPhura: Yeah good call, Ill add a check

NickPhura: I also was setting all admins with read/write access, so I've fixed that to just be sysadmin and bcmi

api/src/controllers/put/collection-bcmi.js
NickPhura: Added

NickPhura:
Updated to address all of the current comments

NickPhura:
> 
> 
> Entering invalid characters and then hitting save blows away any valid date in the collection document

This seems to be the case for all of our editable date fields (based on trying to edit a few random records), so this might be better suited for a bug ticket, and then someone can just address all of them at once.

MaxWardle:
Works as expected and looks good on mobile.

marklise:
> > Entering invalid characters and then hitting save blows away any valid date in the collection document
> 
> @marklise This seems to be the case for all of our editable date fields (based on trying to edit a few random records), so this might be better suited for a bug ticket, and then someone can just address all of them at once.
> 
> Added a bug ticket: https://bcmines.atlassian.net/browse/NRPT-277

Inspection, Certificate are just a couple I found when doing spot checks.  I think this should have been addressed in this PR.

PR #505, opened 2020-07-21
NRPT-241: Multi-select add records to collection

NickPhura:
Ticket: https://bcmines.atlassian.net/browse/NRPT-241

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
@@ -8,7 +8,7 @@ import { LoadingScreenService, Utils, StoreService } from 'nrpti-angular-compone
 import { of } from 'rxjs';
 import { catchError, takeUntil } from 'rxjs/operators';
 import { Subject } from 'rxjs/Subject';
-import { Picklists, StateIDs } from '../../../../../common/src/app/utils/record-constants';
+import { Picklists, StateIDs, StateStatus } from '../../../../../common/src/app/utils/record-constants';
 import { ConfirmComponent } from '../../confirm/confirm.component';
 import { RecordUtils } from '../../records/utils/record-utils';
 import { FactoryService } from '../../services/factory.service';
@@ -38,6 +38,9 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
   public collectionTypes = Picklists.collectionTypePicklist;
   public collectionAgencies = Picklists.collectionAgencyPicklist;
 
+  // collection add edit state
+  public collectionState = this.storeService.getItem(StateIDs.collectionAddEdit);
+
   constructor(
     public route: ActivatedRoute,
     public router: Router,
@@ -52,6 +55,8 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
 
   ngOnInit() {
     this.loadingScreenService.setLoadingState(true, 'main');
+
+    this.setOrRemoveCollectionAddEditState();
 
     this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
       this.isEditing = res.breadcrumb !== 'Add Collection';
@@ -74,6 +79,22 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
 
       this._changeDetectionRef.detectChanges();
     });
+  }
+
+  /**
+   * Sets the initial collectionAddEdit state, or removes it from the store if it is invalid.
+   *
+   * @memberof MinesCollectionsAddEditComponent
+   */
+  setOrRemoveCollectionAddEditState() {
+    const tempCollectionAddEditState = this.storeService.getItem(StateIDs.collectionAddEdit);
+    if (tempCollectionAddEditState) {
+      if (tempCollectionAddEditState.status === StateStatus.invalid) {
+        this.storeService.removeItem(StateIDs.collectionAddEdit);
+      } else {
+        this.collectionState = tempCollectionAddEditState;
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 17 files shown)


angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.html
```diff
@@ -50,6 +51,31 @@ <h1 id="name" class="m-0">{{ (mine && mine.name) || '-' }}</h1>
       [filters]="filters">
     </search-filter-template>
   </section>
+
+  <section *ngIf="collectionAddEditState">
```
dhlevi, line 55: When you select a row, it doesn't seem to change anything... `collectionAddEditState` doesn't appear to be flagged... but if you change tabs, suddenly the button appears! Looks like we need a change detection here, or more likely, this line:
`public collectionAddEditState = this.storeService.getItem(StateIDs.collectionAddEdit);`
Is setting it's default incorrectly

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
```diff
@@ -413,6 +466,37 @@ export class MinesRecordsListComponent implements OnInit, OnDestroy {
...
+      }
+    }
+  }
+
+  /**
+   * Submit adding records to collection.
+   *
+   * @memberof MinesRecordsListComponent
+   */
+  submitAddEditCollectionRecords() {
```
dhlevi, line 492: Method is fine, but I think we'll need a discussion about the behavior with this button click. You're just following the spec so no criticism here. But. When you come into the form, you have no way of knowing if the click will route you to the `add collection` state, or the `edit collection` state. You have no identifier that a collection has already been 'selected' if you came in from the collection screen, and no way to clear a collection if you wanted to.

Because the breadcrumbs are not helpful, it only makes it worse as you have no idea, at a glance, how you got to this screen if you're coming from a collection. It's very un-intuitive and definitely needs more UI cues as to what is going on. I was confused and I knew what I was doing.

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
```diff
@@ -413,6 +466,37 @@ export class MinesRecordsListComponent implements OnInit, OnDestroy {
     );
   }
 
+  /**
+   * Cancel adding records to collection.
+   *
+   * @memberof MinesRecordsListComponent
+   */
+  cancelAddEditCollectionRecords() {
+    const shouldCancel = confirm(
```
dhlevi, line 475: Clicking cancel always routes you to the `add\edit collection` screen, rather than just clearing the "selected" collection and checked records. If you came in from the Mine record and just wanted to create a net-new collection, this would be unexpected!

angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
```diff
@@ -216,6 +251,33 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
     this.myForm.get('collectionRecords').markAsDirty();
   }
 
+  /**
+   * Add records to the collections array of associated records.
+   *
+   * @memberof MinesCollectionsAddEditComponent
+   */
+  addRecordsToCollection() {
```
dhlevi, line 259: When you add a record to a collection, the state is preserved. So navigating away from the form and coming back will restore your selections. The only way to "clear" a selection set is to click cancel, which takes you to the `add\edit collection` screen. I think we should be able to clear the selection states without basically restarting the whole process.

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.html
KitArmstrong: Further do Dylan's comment, I don't actually think the button should appear.  If I navigate to the Records page and select a few records the buttons doesn't appear (correct, I should be using the actions menu).  Then if I navigate to the Collections tab and back to the Records tab I see the button "Add to Collection." When I click it the records are added to a blank collection that doesn't exist. 

I think when selecting records first any action taken on them must come form the Actions menu.

dhlevi: Yeah, fair point. I think this just speaks to the confusion on the screens around the expected behaviour.

NickPhura: This should hopefully be fixed now.  
The issue was that I was allowing the state to get updated when you weren't already editing a collection, which isn't a workflow that makes sense.  So now, the records list page should not be able to set the state from scratch, and only if someone else has already created it for them (ie, the collections add/edit page).

angular/projects/admin-nrpti/src/app/mines/mines-records-list/mines-records-list.component.ts
NickPhura: This should be fixed.  If you cancel, it should send you back to the add/edit collection page, with whatever values you started with originally.  Before, I was still sending along the latest records that you added, which isn't right.

NickPhura:
Started replying to some of your comments, but I think it might just be best to do a re-review, as some of the comments were base on underlying errors I had, so some of the bahaviour you had concerns about shouldn't exist now (as originally intended).

I still haven't done anything with the breadcrumbs.  I think there is still some discussion around that happening here: https://bcmines.atlassian.net/browse/NRPT-245

I haven't added a popup message when the user returns to the collections add/edit page, mostly because of time, but I did add a records count bit of text to the records list page, under the green add button.

NickPhura:
> Right now the state of selected records is always being maintained, which means if I choose some records but don't click 'Add to Collection' then navigate away, when I eventually return to the Records tab it still appears as if I am working on the collection. I think we will need a way to determine if the users is following the flow or breaks it and then clear the state.

Noticed I still need to address this, so ill look at that

NickPhura:
Okay, just pushed a final commit, which just had some quality of life updates (making variable names consistent between components, etc).

NickPhura:
> Thanks for cleaning up the old alert() I left in there.

No worries, I only noticed it because I was running into an async unit test failing, which doesn't report itself correctly, so I ended up scouring lots of the mines pages for anything that might have caused it (was annoying and took a while).  That alert wasn't hurting anything, but figured I'd clean it up while I was working.  That is also why I have some seemingly unrelated 'async()' additions to some tests, trying to track down the mystery error.

NickPhura:
> Approving, but still might want to look at adjusting checkbox colour.

Yeah I am with you on the colors being not great, but I wonder if this is better suited for its own ticket, and then someone can make the scss changes for all of the checkboxes/sliders across the site.  Ill make a ticket for it at least, since that color change is probably a nice sized bug ticket that someone could take.

PR #506, opened 2020-07-21
NRPT-199: Common search style updates

dhlevi:
Bolding all labels, padding on date filter inputs for better spacing on staggered forms

See ticket [NRPT-199](https://bcmines.atlassian.net/browse/NRPT-199)


dhlevi:
> I know we wanted to try and use Bootstrap classes for anything we could. Might be able to use them here using their spacing classes.
> 

If only it were so simple *shakes fist at bootstrap*

For padding, you use the p<side>-<size> classes, which use sizes of 1 through 5. These work off of multipliers to the `$spacer` value. That means that you have to find a size/spacer combo that works... which is great if your `<div>` lives in a column with not other spacing internally. If you need to set a padding or margin that would be outside of those exact multipliers, it won't work... it'll always be off by 4-5 pixels, like so:

image

We can get around that by adding our spacers or custom values for padding/margins... but that seems like doing the same job elsewhere?

PR #510, opened 2020-07-22
NRPT-167: Autocomplete multiselect fixes

dhlevi:
Some fixes to correct issues with the multiselect/autocomplete dropdown boxes in the search component. 

Corrected the text input so pressing 'enter' with a partial match will select the first item on the list (you don't have to fill out the whole word completely anymore). After selection, the "options" list is cleared, so the list will display again as expected.

There are two approaches to correcting the issue of having the text retained after entering/selecting an option. The issue is that the autocomplete box must be cleared between searches. Retaining the text will prevent the selection box from displaying or any further autocompletes.

To get around this, we have two options. If you have the `useChips` parameter set to `false` (the default), your selection text will be applied to the placeholder text of the input box:
image
The one caveat to this approach is that you can't easily see all of your selections once you have several. This issue exists on a regular text box as well of course.

As a secondary option, if you have the `useChips` parameter set to `true`, the placeholder text retains its original value, and material chips are added below the input box:
image

Currently all multiselects are set to the defeault (useChips = false) but it's a pretty easy swap if/where preferred.

See ticket [NRPT-167](https://bcmines.atlassian.net/browse/NRPT-167)


dhlevi:
I've changed the default in the FilterObject class to `useChips = true`. Every autocomplete dropdown should now be using chips! Likely we'll have to do some testing and review to make sure it makes sense... it's pretty easy to change the default or specify some as chips and others as not.

PR #515, opened 2020-07-27
NRPT-198: Link records to collection

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-198

Adds the ability to select records from the records list and add them to a new or existing collection.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
@@ -2,6 +2,7 @@ import { CdkDragDrop, moveItemInArray } from '@angular/cdk/drag-drop';
 import { ChangeDetectorRef, Component, OnDestroy, OnInit } from '@angular/core';
 import { FormArray, FormControl, FormGroup } from '@angular/forms';
 import { ActivatedRoute, Router } from '@angular/router';
+import {Location} from '@angular/common';
 import moment from 'moment';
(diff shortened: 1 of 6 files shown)


mtCarto:
Looks good, can add records to new or existing collections. 
Did notice a small issue when adding to an existing collection the records don't initially show up in the collection view. On refresh they do though.

KitArmstrong:
> Looks good, can add records to new or existing collections.
> Did notice a small issue when adding to an existing collection the records don't initially show up in the collection view. On refresh they do though.

Just tested 5 times and did not receive this. Will keep an eye out and make a ticket if it becomes an issue.

KitArmstrong:
> 👍 Looks good, works as expected.
> 
> Not an issue with the the work you've done at all, but the process is confusing; while the actions themselves are sensible and intuitive enough, the screens themselves are not. It's very strange when you select the 'add to existing' option that you're just in the usual collection panel with no real indicator of what you're doing there. You can click "add to collection" and add more records while you're in there too, which kind of gets you lost in the process (it all works fine, but some kind of visual indicator as to your current state would help).

I agree. I do not think the way it currently behaves is intuitive at all. Screens are being reused too many times and it makes the navigation hard to follow and super easy to get out of sync.  Will see what the feedback is during further testing.

dhlevi:
> > Looks good, can add records to new or existing collections.
> > Did notice a small issue when adding to an existing collection the records don't initially show up in the collection view. On refresh they do though.
> 
> Just tested 5 times and did not receive this. Will keep an eye out and make a ticket if it becomes an issue.

I didn't see this happening either... could be just a weird state you can get into when jumping between forms?

PR #516, opened 2020-07-27
NRPT-297: Permit Updates - Model and Migration

dhlevi:
Remodel for PermitBCMI flavour model, removal of PermitAmendment and flavour models. Migration included to convert existing Permit and PermitAmendment records to the new model by flattening PermitAmendment records by their list of documents (each doc becomes a Permit/PermitBCMI record).

WARNING: This will break the Core importer, so might be worth waiting to merge until after the Core importer has been updated before merging (or merge them close together? Not a fan of pushing in code that breaks functionality).

Note: Because of the restructuring, it might also be more worth while to clean out all BCMI core imported records and re-run the importer, rather than running the migration script to restructure the existing data (this stuff isn't in prod yet so makes sense to just flush). If that's considered a better approach we can remove the script from here.

See Ticket [NRPT-297](https://bcmines.atlassian.net/browse/NRPT-297)


api/migrations/20200724192045-flattenPermitModel.js
```diff
@@ -0,0 +1,256 @@
...
+  await nrpti.remove({ _id: new ObjectId(amendment.flavour._id) });
+  await nrpti.remove({ _id: new ObjectId(amendment.record._id) });
+
+  // we need to return something to declare as the "OG" if this is creating
+  // an OGP. May as well return the last thing created.
+  return ogGuid;
+}
+
+async function createPermitFlavour(nrpti, permitFlavour, originalPermitRef, amendment, doc) {
+  let PermitBCMI = require('../src/models/bcmi/index.js').permitBCMI;
```
KitArmstrong, line 142: Nit: Use destructing instead  ``` const { PermitBCMI } = require('../src/models/bcmi/index.js')```

api/src/models/bcmi/permit-bcmi.js
```diff
@@ -1,25 +1,46 @@
...
   {
+    // NRPTI Record boilerplate
     _schemaName: { type: String, default: 'PermitBCMI', index: true },
     _sourceRefId:  { type: String, default: null, index: true },
     _epicProjectId: { type: 'ObjectId', default: null, index: true },
-
     read: [{ type: String, trim: true, default: 'sysadmin' }],
     write: [{ type: String, trim: true, default: 'sysadmin' }],
-
+    recordName: { type: String, default: '' },
```
KitArmstrong, line 10: Should any of these new fields be included on the master model? I am not sure if they should or not, but normally I think the models align fairly closely.

dhlevi: Yeah, I wasn't sure so I figured it would come up in the review. Easy to match them up though so I can add them for master/flavour

api/migrations/20200724192045-flattenPermitModel.js
dhlevi: updated: 
```
const { permitBCMI: PermitBCMI } = require('../src/models/bcmi/index');
const { permit: Permit }  = require('../src/models/master/index');
```

api/src/models/bcmi/permit-bcmi.js
dhlevi: I've added the attributes to the Permit master and the LNG flavour. Noticed the PermitBCMI model was missing in common as well so I added that in for completeness.

PR #520, opened 2020-07-28
NRPT-265 Adds BCMI Record Detail View

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-265

Adds a simple record view that is used in the Mines/BCMI area when viewing records from the record list.


angular/projects/admin-nrpti/src/app/mines/mines-records-rows/mines-records-table-row.component.spec.ts
```diff
@@ -1,33 +1,36 @@
...
-    const { component, fixture } = testBedHelper.createComponent(false);
-
-    component.rowData = {};
-
-    fixture.detectChanges();
-
-    expect(component).toBeTruthy();
-  });
-});
+ // TODO: see if test can be fixed. Using routerLink creates errors with blank paths. See bug: https://github.com/angular/angular/issues/27674
```
KitArmstrong, line 1: I had to adjust the routing which resulted in this test suffering from a routing bug that some others were. Added a note at the top like the others.

api/src/controllers/search.js
```diff
@@ -42,6 +43,8 @@ let generateExpArray = async function (field, logicalOperator = '$or', compariso
           return convertValue(element);
         });
         return { [logicalOperator]: [{ [item]: { $in: arrayExp } }] };
+      } else if (!Array.isArray(entry) && comparisonOperator === '$in') {
```
KitArmstrong, line 46: I searched and couldn't find the `_in` value being used anywhere else, so I don't think making this change will cause any issues.

angular/projects/admin-nrpti/src/app/mines/mines-records-rows/mines-records-table-row.component.ts
```diff
@@ -70,52 +71,7 @@ export class MinesRecordsTableRowComponent extends TableRowComponent implements
    * @memberof MinesRecordsTableRowComponent
    */
   goToDetails() {
-    switch (this.rowData._schemaName) {
```
dhlevi, line 73: After working on #521 I cannot begin to express how much happiness I get from this change.

angular/projects/admin-nrpti/src/app/mines/mines-records-detail/mines-records-detail.component.html
```diff
@@ -0,0 +1,82 @@
...
+  <section class="mb-3 pb-3 border-bottom">
+    <div class="row">
+      <div class="mb-2 col-12 col-sm-8">
+        <h2 class="border-0 mb-0">Record Information</h2>
+        <span class="grey-subtext">{{ lastEditedSubText }}</span>
+      </div>
+      <div class="col-12 col-sm-4">
+        <div class="d-flex flex-column">
+          <mat-slide-toggle
+            name="publishLng"
```
dhlevi, line 19: Should probably rename/id the component `publishBcmi`, rather than LNG

angular/projects/admin-nrpti/src/app/mines/mines-records-detail/mines-records-detail.component.html
```diff
@@ -0,0 +1,82 @@
+<main class="container-fluid-padding">
```
dhlevi, line 1: Seems to be missing the "date" attribute display as per the AC? Not sure which "date" is supposed to be here though... kinda unclear in the ticket.

KitArmstrong: It actually shows the date on line 14. If there is an updated date it will say "Updated on blah" if it is missing then it will use the created date and say "Created on blah."  This follows how the date is displayed in other areas, but if it isn't obvious enough I could change it into a field like type/agency.

angular/projects/admin-nrpti/src/app/mines/mines-records-detail/mines-records-detail.component.html
KitArmstrong: Fixed.

angular/projects/admin-nrpti/src/app/mines/mines-records-detail/mines-records-detail.component.html
dhlevi: No that's probably fine. The design has it as a defined field... but for a create/update date a user should not be able to change it so having it the way you've done makes sense.

PR #525, opened 2020-07-30
NRPT-116: mem-admin doc import, bcmi flavours

dhlevi:
Adding in a migration to import docs from MEM-ADMIN and create the related collections. In addition, BCMI flavours are missing for some types that will need one, so I've added them.

Note: Talking with @marklise earlier and we were discussing changing the s3 upload from using the environment vars for the s3 bucket to maybe using the API. Putting the PR up here pre-emptively to get a review over the migration and added flavours, but the migration may change if I swap out the aws calls for direct API calls. Otherwise, be sure to set your env vars for your s3 bucket to test.

See ticket [NRPT-116](https://bcmines.atlassian.net/browse/NRPT-116)


api/migrations/20200709183253-loadMemDocs.js
```diff
@@ -0,0 +1,336 @@
...
+                console.error('#######################################');
+                errors += 1;
+              }
+            }
+          }
+          console.log(`Creating NRPTI collection for ${collection.displayName}`);
+          // Master, Meta and Documents for this collection are all created.
+          // Now, create a NRPTI collection and shove the docs into it!
+          let bcmiCollection = new CollectionBCMI();
+          bcmiCollection._master = new ObjectID(nrptiMine._id);
```
KitArmstrong, line 122: Have constants for these values.

api/migrations/20200709183253-loadMemDocs.js
```diff
@@ -0,0 +1,336 @@
...
+          bcmiCollection.name = collection.displayName;
+          bcmiCollection.date = new Date();
+          bcmiCollection.project = new ObjectID(nrptiMine._id);
+          bcmiCollection.type = collection.type;
+          bcmiCollection.agency = collection.isForMEM ? 'EMPR' : collection.isForEAO ? 'EAO' : 'ENV';
+          bcmiCollection.records = allNewDocs;
+          bcmiCollection.addedBy = 'nrpti';
+          bcmiCollection.datePublished = new Date();
+          bcmiCollection.publishedBy = 'nrpti';
+          bcmiCollection.isBcmiPublished = true;
```
KitArmstrong, line 134: Just confirming we want this auto-published?

dhlevi: I assumed so, as they're published in mem-admin, but we can always leave that as an exercise for an admin if it's safer

PR #527, opened 2020-07-31
NOBUG: Record model, API refinements

dhlevi:
Optional NOBUG PR, can be ignored if there's no desire to regression test or include these changes. Just a simple set of refinements to the NRPTI "Record" model classes, and UI/API side cleanup of the large `if` and `case` statements. Trims out a few thousand lines of mainly duplicate code, and is intended to improve maintainability if we need to add more record types in the future. Loads of room for further enhancement, particularly around making common components for the UI/Flavours that can build themselves from the model (they're all so similar it should be pretty easily doable).

Updates:
- UI Factory: Replaces the `create*` and `edit*` with a single `writeRecord` function
- UI Record List Component: Removed large case with a single `this.router.navigate(['records', item, 'add']);` and updated passed in items to be the values we actually needed to use.
- UI Record Row Component: Update onClick and edit clicks to remove duplicated case statement (still need at least one here unfortunately)
- UI RecordUtils: Replaced large `case` in `getRecordModelInstance` with a reflective instantiator.
- Common Models: Added a Record Model interface class containing all common attributes. Added a Record Model abstract class inheriting that interface. Updates all Model classes to extend the abstract class. Can be further enhanced for flavour specific abstracts, or update to include methods for save, reload, etc.
- API Record Controller: Removed large `if` and `case` statements, replaced with a const and an Object.Keys check.


angular/projects/admin-nrpti/src/app/records/records-list/records-list.component.ts
```diff
@@ -446,67 +446,7 @@ export class RecordsListComponent implements OnInit, OnDestroy {
   }
 
   add(item) {
-    switch (item) {
```
KitArmstrong, line 449: This is so much nicer to see.

angular/projects/admin-nrpti/src/app/records/records-rows/records-table-row.component.ts
```diff
@@ -34,131 +34,60 @@ export class RecordsTableRowComponent extends TableRowComponent implements OnIni
...
-        this.router.navigate(['records', 'annual-reports', this.rowData._id, 'edit']);
-        break;
-      default:
-        break;
-    }
+    this.routeToRecordsPage('edit');
+  }
+
+  private routeToRecordsPage(type) {
+    this.router.navigate(['records', this.getSchemaRoute(this.rowData._schemaName), this.rowData._id, type]);
```
KitArmstrong, line 91: Very nice imporvement.

angular/projects/admin-nrpti/src/app/records/utils/record-utils.ts
```diff
@@ -381,4 +235,14 @@ export class RecordUtils {
       }
     }
   }
+
+  static getReflectiveInstance(context: any, name: string, ...args: any[]) {
```
KitArmstrong, line 239: Does this need to be static? I didn't see it used anywhere else, but could have missed it. Or, maybe it is nice to just have available?

dhlevi: Yeah, it doesn't need to be static, but I figured it could get use elsewhere whenver some similar action needs to be done.

PR #529, opened 2020-08-05
NRPT-260: Adding/Editing Records for a Mine

dhlevi:
Added screens and functionality for adding/editing records within the context of a mine. Includes a merge from existing branch from @mtCarto for the bulk of the Add code.

Note: From the AC for NRPT-197, there is a request to auto-generate and auto-publish a record when it is added to a collection. This is not in place due to a number of technical reasons which should likely be addressed in a future ticket as the changes required go beyond the scope of simply adding/editing records for a mine.

See tickets [NRPT-260](https://bcmines.atlassian.net/browse/NRPT-260) and [NRPT-197](https://bcmines.atlassian.net/browse/NRPT-197)


dhlevi:
Whoops, forgot to squash before PRing. Apologies about the lengthy commit history!

angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    const record = {};
+
+    this.myForm.get('recordName').dirty && (record['recordName'] = this.myForm.get('recordName').value);
+    this.myForm.get('recordDate').dirty &&
+      (record['dateIssued'] = this.utils.convertFormGroupNGBDateToJSDate(this.myForm.get('recordDate').value));
+    record['recordType'] = this.myForm.get('recordType').value;
+    this.myForm.get('recordAgency').dirty && (record['issuingAgency'] = this.myForm.get('recordAgency').value);
+
+    // lookup appropriate schemaName from from type value
```
KitArmstrong, line 255: "from from"

angular/projects/admin-nrpti/src/app/services/factory.service.ts
```diff
@@ -350,6 +350,139 @@ export class FactoryService {
...
+   * Edit a mine record.
+   *
+   * @param {*} record object containing mine record values to update into a record item.
+   * @returns {Observable<object>}
+   * @memberof FactoryService
+   */
+  public editMineRecord(record: any): Observable<object> {
+    const outboundObject = {};
+
+    // theres a bunch missing?
```
KitArmstrong, line 363: Are there?

angular/projects/admin-nrpti/src/app/services/factory.service.ts
```diff
@@ -350,6 +350,139 @@ export class FactoryService {
...
+        break;
+      case 'Warning':
+        outboundObject['warnings'] = [record];
+        break;
+      default:
+        outboundObject['records'] = [record];
+        break;
+    }
+
+    console.log(outboundObject);
```
KitArmstrong, line 424: Log

angular/projects/admin-nrpti/src/app/services/factory.service.ts
```diff
@@ -350,6 +350,139 @@ export class FactoryService {
...
+
+  /**
+   * Delete a mine record.
+   *
+   * @param {string} recordId _id of the record to delete.
+   * @param {string} model schema of record to delete
+   * @returns {Promise<any>}
+   * @memberof FactoryService
+   */
+  // todo verify need model specified
```
KitArmstrong, line 481: Is this still a todo?

api/src/controllers/post/dam-safety-inspection.js
```diff
@@ -89,7 +89,7 @@ exports.createMaster = function (args, res, next, incomingObj, flavourIds) {
 
   // set data
   incomingObj.recordName && (damSafetyInspection.recordName = incomingObj.recordName);
-  damSafetyInspection.recordType = 'DamSafetyInspection';
+  damSafetyInspection.recordType = 'Dam Safety Inspection';
```
KitArmstrong, line 92: Nice catch, I noticed this during the last demo.

angular/projects/admin-nrpti/src/app/services/factory.service.ts
dhlevi: Well, there was! That comment doesn't make as much sense anymore.

However, does bring up a note that some of the types in the list do not have BCMI flavours and therefore will not work. Those types should either have flavours added, or be removed from the list if they are not needed for BCMI.

angular/projects/admin-nrpti/src/app/services/factory.service.ts
dhlevi: Nope, fixed

angular/projects/admin-nrpti/src/app/services/factory.service.ts
dhlevi: Removed

angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
dhlevi: Fixed

dhlevi:
> * I think we need a way to add a permit amendment.  Whether it is a new dropdown that appears if permit is select or if permit amendment appears as a type and just handles it all in the back, I think the option needs to be there.

In progress. Looking into the best way to implement in the add/edit UI now.

> * The Add Record screen has the title "Edit Record."

Fixed!

> * Some record types throw an error when saving.  `ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'subscribe' of null TypeError: Cannot read property 'subscribe' of null at MinesRecordsAddEditComponent.<anonymous> (mines-records-add-edit.component.ts:309)`

Couldn't reproduce, but I made a few other changes so may have been inadvertently fixed? Likely related to types without BCMI flavours, which have been removed for now until those types are updated.

> * Sometimes get a generic error when creating that just says "Failed to create record" with no indication why. Name = "Testing", Date = "2020-08-05",   Type = "Agreement", Agency = "EAO"

Couldn't reproduce, but I made a few other changes so may have been inadvertently fixed?

> * Doesn't appear to be a way to add documents to a record when creating it

Fixed!

> * Once you finish editing it directs you to the details view. The only way to get back to the records list is the use the breadcrumbs at the top.  Should there be some other action button?

Yeah, could be a good idea. Looks like that's pretty consistent in all of our Details screens. I'll leave it as-is for now to be consistent, but it would be a good idea to get a recommendation from QA and make a ticket if appropriate.

dhlevi:
> This code looks very similar to #538. Maybe in the future we will have time to consolidate this code.

It's probably more or less identical. Good idea to flag it as technical debt to merge into something shared in the future.

danieltruong:
> > This code looks very similar to #538. Maybe in the future we will have time to consolidate this code.
> 
> It's probably more or less identical. Good idea to flag it as technical debt to merge into something shared in the future.

I'll make a ticket

PR #533, opened 2020-08-07
NRPT-276: Adds issued to to permit

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-276

Most of the work had already been completed. The only thing missing was the IssuedTo field.


KitArmstrong:
> Looks good, couple of quick things (if they're relevant, not sure if they are yet):
> 
> Should this attribute be added to the BCMI/NRCED/LNG flavour models?
> Need to update the UI model in the Common package (And flavours where needed)
> There's no way to view/edit, so we'll need to update the Permit UI.
> 
> If we don't need to update those parts yet, ignore me!

I added the property to the flavours and updated the UI models.  Will leave the edit screens right now as this was just to fix the display in the All Records list.

PR #534, opened 2020-08-10
NRPT-319: Document publish fixes

dhlevi:
Changes to allow documents to be published on the creation of a new record, when a record is published, or when a document is updated. Additionally, a couple of other small fixes to issued identified while fixing the core issue.

Updated the businessLogicManager `isDocumentConsideredAnonymous` function so a new record without documents or a record that does not have an `issuedTo` attribute could be public. Previously the only documents that could ever be published had a very thin criteria that expected some values that would not be consistent with all record types.

Also added a step when publishing a record to go through and publish any associated docs. This will call the `isDocumentConsideredAnonymous` function as well, for consistency and to prevent publishing a document that should not be public.

Additionally, made a small fix in the document controllers `POST` method to check the supplied record ID was valid. It was possible to completely kill the service (or silently fail in some cases) by passing in an invalid ObjectID.

Finally, records were passing in an invalid objectID for the master when creating new records due to some parsing issues when fetching the result. This would allow the document to be saved, but ignored any permissions and always forced the documents to be anonymous.

See ticket [NRPT-319](https://bcmines.atlassian.net/browse/NRPT-319)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/records/administrative-penalties/administrative-penalty-add-edit/administrative-penalty-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/records/administrative-penalties/administrative-penalty-add-edit/administrative-penalty-add-edit.component.ts
@@ -406,11 +406,18 @@ export class AdministrativePenaltyAddEditComponent implements OnInit, OnDestroy 
     if (!this.isEditing) {
       this.factoryService.writeRecord(administrativePenalty, 'administrativePenalties', true).subscribe(async res => {
         this.recordUtils.parseResForErrors(res);
+        let _id = null;
+        if (Array.isArray(res[0][0].object)) {
+          _id = res[0][0].object.find(r => r._schemaName === 'AdministrativePenalty')._id;
+        } else {
+          _id = res[0][0].object._id;
+        }
+
         const docResponse = await this.recordUtils.handleDocumentChanges(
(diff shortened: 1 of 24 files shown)


angular/projects/admin-nrpti/src/app/records/administrative-penalties/administrative-penalty-add-edit/administrative-penalty-add-edit.component.ts
```diff
@@ -406,11 +406,18 @@ export class AdministrativePenaltyAddEditComponent implements OnInit, OnDestroy
     if (!this.isEditing) {
       this.factoryService.writeRecord(administrativePenalty, 'administrativePenalties', true).subscribe(async res => {
         this.recordUtils.parseResForErrors(res);
+        let _id = null;
+        if (Array.isArray(res[0][0].object)) {
+          _id = res[0][0].object.find(r => r._schemaName === 'AdministrativePenalty')._id;
```
KitArmstrong, line 411: Do we have constants available for these?

dhlevi: We do for the whole collection of Schema types, but not a constants list of them individually. I can add one if it's preferred to just strings here.

PR #535, opened 2020-08-10
NRPT-283: Update collection advanced filters

dhlevi:
Adds remaining filters to the collections advanced filters. Implements Record filter.

See ticket [NRPT-283](https://bcmines.atlassian.net/browse/NRPT-283)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-list-resolver.ts
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-list-resolver.ts
@@ -44,6 +44,36 @@ export class MinesCollectionsListResolver implements Resolve<Observable<object>>
       and['project'] = params.mineId;
     }
 
+    if (params.type) {
+      or['type'] = params.type.join(',');
+    }
+
+    if (params.agency) {
+      or['agency'] = params.agency.join(',');
+    }
+
+    if (params.bcmiTabType) {
+      if (!or['type']) {
+        or['type'] = '';
+      }
+
+      if (params.bcmiTabType.includes('Authorizations')) {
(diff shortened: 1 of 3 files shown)


angular/projects/admin-nrpti/src/app/mines/mines-collections-list-resolver.ts
```diff
@@ -44,6 +44,36 @@ export class MinesCollectionsListResolver implements Resolve<Observable<object>>
...
+    if (params.type) {
+      or['type'] = params.type.join(',');
+    }
+
+    if (params.agency) {
+      or['agency'] = params.agency.join(',');
+    }
+
+    if (params.bcmiTabType) {
+      if (!or['type']) {
```
KitArmstrong, line 56: With the check on line 72, does this need to be here?

dhlevi:
> Noticed something odd. I don't think this is because of your code. It's likely bad data on my end. Those records that are published and appearing have isBcmiPublished set to false but its flavour does have the public role in its read.

Yep, noticed the same thing. The isBcmiPublished flag was actually not being set in a number of locations when a record was published. These have been updated where caught in the API side (part of another ticket). I'll go through and test a few others and if needed created a follow up ticket to catch any missing put's that are not setting the flag where appropriate

angular/projects/admin-nrpti/src/app/mines/mines-collections-list-resolver.ts
dhlevi: Yeah, looks weird but is required. Basically I create a 'stub' or object and populate if needed. It's possible it might be empty by the time the checks are done, so in that case I just make sure to delete the type ref so it isn't going to inadvertently search for null types and return incorrect results

PR #537, opened 2020-08-11
NRPT-57: Allows multiple NRIS imports

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-57

Stops duplication when the NRIS importer is run more than once. Also includes a migration to add the source reference ID for the NRIS records as there was an issue that was saving them as `null`.

### DIFF
--- a/api/src/models/bcmi/permit-bcmi.js
+++ b/api/src/models/bcmi/permit-bcmi.js
@@ -27,12 +27,8 @@ module.exports = require('../../utils/model-schema-generator')(
     dateIssued: { type: Date, default: null },
     authorizedEndDate: { type: Date, default: null },
     description: { type: String, default: '' },
-    // Document ref to meta record and physical document on the object store
-    amendmentDocument: {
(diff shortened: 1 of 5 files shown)


api/src/models/bcmi/permit-bcmi.js
```diff
@@ -27,12 +27,8 @@ module.exports = require('../../utils/model-schema-generator')(
     dateIssued: { type: Date, default: null },
     authorizedEndDate: { type: Date, default: null },
     description: { type: String, default: '' },
-    // Document ref to meta record and physical document on the object store
-    amendmentDocument: {
-      _sourceRefId: { type: String, default: null },
-      documentName: { type: String, defualt: '' },
-      documentId: { type: 'ObjectId', default: null }
-    },
+    documents: [{ type: 'ObjectId', default: [], index: true }],
```
KitArmstrong, line 30: Fixed to match the master.

PR #539, opened 2020-08-11
NRPT-269: Enforce record in one collection

dhlevi:
Updates to UI to disable the select/check boxes if a record is a member of a collection. Remove the "hasCollection" filter due to implementation difficulties (not possible with our model in MongoDB 3.6)

Update to API search to append collections to records, if they have them (with a note about changes we can do with MongoDB 4.x to improve the process)

Updates to collections post/put to ensure records passed in do not contain any records that are already a member of another collection

See ticket [NRPT-269](https://bcmines.atlassian.net/browse/NRPT-269)


dhlevi:
> I wonder if it should be removed when a record belongs to a collection?

Yeah, makes sense. Easy enough to just hide them rather than disable. I'll make that change and if QA disagrees we can swap it back.

dhlevi:
> Wonder if we could just remove those records completely from the results of the record list. That would be pretty clear that the record is not available.

Definitely another alternative, but then the lists functionality as a descriptor for all records within the context of a mine isn't really true. The screen does triple-duty right now by being:

a.) The searchable list of all Mine related records
b.) A method for creating new records
c.) A method for associating records to a collection

So... I agree we should remove them when we want to associate, but that's just more logic that winds up going into this screen that's state dependent. Which is going to get buggy and ugly fast. If we implement a secondary screen for that function though, it's definitely the best approach to just exclude them from the results.

marklise:
Small issue, when clicking the checkbox cell area I would have expected the entire cell be clickable, not just the checkbox itself.  Also the checkbox itself isn't centered in the cell. See screenshot showing the area which I think should respond to enabling/disabling the check:

image

dhlevi:
Updated so the cell is now clickable as well. I tried setting the element to center but the defaults on the angular component were getting in the way. I can look into this later.

PR #540, opened 2020-08-12
NRPT-278: Adding descriptive text

dhlevi:
Adds some descriptive text to the mine add/edit pages description and summary text labels.

See ticket [NRPT-278](https://bcmines.atlassian.net/browse/NRPT-278)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-add-edit/mines-add-edit.component.scss
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-add-edit/mines-add-edit.component.scss
@@ -3,5 +3,5 @@
 .summary-label {
   font-size: small !important;
   font-style: italic;
-  color: gray;
+  color: $font-gray;
 }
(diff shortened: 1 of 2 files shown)


angular/projects/admin-nrpti/src/app/mines/mines-add-edit/mines-add-edit.component.scss
```diff
@@ -1 +1,7 @@
 @import 'assets/styles/components/mine.scss';
+
+.summary-label {
+  font-size: small !important;
+  font-style: italic;
+  color: gray;
```
KitArmstrong, line 6: There are different grays in the theme file. Might want to pull one of those. There might even be one specific for text IIRC.

dhlevi: Swapped out the colour to use the theme `$font-gray`

PR #544, opened 2020-08-12
NRPT-262: Associating mine records.  Fixing unit tests, code cleanup.

marklise:
The main part of this bug was to add associated mine records.  This changes how the project and mine selectors work by assigning the dropdown value to either form element as long as the record is an `nrpti` source.

There is now a service that pulls the epic project list and mine list from the back-end on an interval and stuffs it into the storage service.  The storage service emits this so that we can subscribe to the change and update the UI without requiring a page load.

The task expanded into discovering a slew of issues w.r.t tests and discovering race conditions in running them, and just finding and fixing a whole lot of things.  I found sometimes they would fail, and sometimes they would pass.  Usually the first run would pass, and the second run would fail (and then alternate pass/fail/pass/fail).  I could split them out but there are so many changes and most are originating from a bad initial copy-paste that got replicated all over the place.

The unit tests we have mostly only bootstrap the test, but in fixing and cleaning them up this should save time for others in making them actually useful.


angular/projects/admin-nrpti/src/app/app.component.ts
```diff
@@ -52,6 +69,90 @@ export class AppComponent implements OnInit {
...
+  }
+
+  // Sets a store service list item, but unshifts a null-based element first.
+  private setStoreServiceItem(key, list) {
+    // First unshift an item into the list.
+    list.unshift({_id: null, name: 'None'});
+
+    // Set the object in the store service
+    // tslint:disable-next-line: prefer-const
+    let newObject = {};
```
KitArmstrong, line 153: Nit - just initialize to the value.
```
const newObject = { [key]: list }
```

marklise: Fixed in 445d09a

danieltruong:
Not sure if this is a problem:

- Go to any record that has a mine guid set.
- Set mine guid to none and save.
- Mine guid is not removed.

PR #550, opened 2020-08-14
NRPT-202: Cleanup ROLES constants

marklise:
Everything sits in the api folder now, possible next improvement is to push the valid roles config into the back-end and query on SPA initialize.

### DIFF
--- a/api/src/utils/constants/misc.js
+++ b/api/src/utils/constants/misc.js
@@ -1,19 +1,11 @@
 exports.SYSTEM_USER = 'SYSTEM_USER';
 
-// Available user roles.
-exports.ROLES = {
-  SYSADMIN: 'sysadmin',
-  LNGADMIN: 'admin:lng',
-  NRCEDADMIN: 'admin:nrced',
-  BCMIADMIN: 'admin:bcmi',
-
-  ADMIN_ROLES: [
-    'sysadmin',
-    'admin:lng',
-    'admin:nrced',
-    'admin:bcmi'
-  ]
-};
+exports.ApplicationRoles = {
+  ADMIN: 'sysadmin',
+  ADMIN_NRCED: 'admin:nrced',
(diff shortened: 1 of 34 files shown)


api/src/utils/task-audit-record.test.js
```diff
@@ -52,11 +52,11 @@ describe('TaskAuditRecord', () => {
 
       expect(mockFindOneAndUpdate).toHaveBeenCalledWith(
         { _id: '456' },
-        { $set: { _id: '456', ...{ param1: 3, param2: 4, read: ROLES.ADMIN_ROLES, write: [ROLES.SYSADMIN] } } },
+        { $set: { _id: '456', ...{ param1: 3, param2: 4, read: utils.ApplicationRoles.ADMIN_ROLES, write: [utils.ApplicationRoles.SYSADMIN] } } },
```
KitArmstrong, line 55: Some of these are still using ADMIN_ROLES which was removed.

api/src/utils/constants/misc.js
```diff
@@ -1,19 +1,11 @@
...
-    'sysadmin',
-    'admin:lng',
-    'admin:nrced',
-    'admin:bcmi'
-  ]
-};
+exports.ApplicationRoles = {
+  ADMIN: 'sysadmin',
+  ADMIN_NRCED: 'admin:nrced',
+  ADMIN_LNG: 'admin:lng',
```
KitArmstrong, line 6: Need to add ADMIN_ROLES back or change everywhere it is used.

marklise: I'll change how it works.

marklise: Fixed in 59a3d4f

api/src/utils/task-audit-record.test.js
marklise: Fixed in 59a3d4f.

PR #568, opened 2020-08-25
NRPT-344: Setting published, search endpoint update

dhlevi:
Update to ensure isBcmiPublished and read role is set when records/collections are updated.

Also added a new search datatype of CollectionDocuments as a public endpoint for BCMI to use to fetch collection documents without requiring a check on the master _id values.

Additionally (and unrelated), made a fix to the hasCollection aggregate section. The hasCollection filter could cause the search to fail (and the api to crash) if you included a text search component. This is because an `$addField` step was created before the `$match: {$text}}` which mongodb does not allow in an aggregate. I've updated the logic to create the addField and match check after the main match processing is done.

See ticket [NRPT-344](https://bcmines.atlassian.net/browse/NRPT-344)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
@@ -280,10 +280,10 @@ export class MinesRecordsAddEditComponent implements OnInit {
       record[schemaString]['typeCode'] = this.myForm.get('typeCode').value;
     }
     if (this.myForm.get('recordPublish').dirty && this.myForm.get('recordPublish').value) {
-      record['addRole'] = 'public';
+      record['isBcmiPublished'] = true;
       record[schemaString]['addRole'] = 'public';
     } else if (this.myForm.get('recordPublish').dirty && !this.myForm.get('recordPublish').value) {
-      record['removeRole'] = 'public';
+      record['isBcmiPublished'] = false;
       record[schemaString]['removeRole'] = 'public';
     }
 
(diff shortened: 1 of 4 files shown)


angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
```diff
@@ -280,8 +280,10 @@ export class MinesRecordsAddEditComponent implements OnInit {
       record[schemaString]['typeCode'] = this.myForm.get('typeCode').value;
     }
     if (this.myForm.get('recordPublish').dirty && this.myForm.get('recordPublish').value) {
+      record['addRole'] = 'public';
```
KitArmstrong, line 283: Should the master also change? I thought that for the BCMI related displays it would just focus on the BCMI flavour.

api/src/controllers/search.js
```diff
@@ -278,6 +278,18 @@ let searchCollection = async function(
...
   }
 
+  // has collection filter
+  // For this filter to work, we need to create a field and re-create the hasCollection match
+  // if we add to the initial match criteria though, it will break the initial match. So we
+  // flag it for addition later and remove the check from the "or" variable
+  let hasCollectionTest = false;
+  let hasCollection = null;
+  if (or && Object.prototype.hasOwnProperty.call(or, 'hasCollection')) {
+    hasCollectionTest = true;
```
KitArmstrong, line 288: Remove?

angular/projects/admin-nrpti/src/app/mines/mines-records-add-edit/mines-records-add-edit.component.ts
dhlevi: Actually this can be removed. Publishing the master is ignored (the code in the API specifically prevents it).

api/src/controllers/search.js
dhlevi: No, this is in to do a pre-check to see if we have a hasCollection filter. If we do, we need to remove it from the "or", otherwise the query to the DB will never return any results. We don't actually have a column for this filter so it has to be added dynamically.

we wouldn't have to do this step, but If we add the field before the match pipeline, the query will fail when you include text searches, because MongoDB aggregates require any $text pipeline to be the very first thing that runs. (Try a text search in test now, and you'll notice it fails)

PR #574, opened 2020-08-27
NRPT-299: Add public search redact for issuedTo

dhlevi:
Add a step in the search endpoint to check the issuedTo values for public searches. If an issuedTo dateOfBirth is less than 19, redact this info and replace with 'Unpublished'. Data is not modified by the process so the issuedTo individual's data is still kept intact.

Brings up an interesting point though... once enough time has passed the 'issuedTo' individuals information will display as they'll be over 18, so should the check actually be against "dateAdded" and not today? I think so...?

See ticket [NRPT-299](https://bcmines.atlassian.net/browse/NRPT-299)


api/src/controllers/search.js
```diff
@@ -419,6 +419,68 @@ let searchCollection = async function(
     }
   });
 
+  // For read only users, we need to redact out the details
+  // of any individual where the birthdate is null or the individual
+  // is less then 19 years old. First step to do this is calculate their
+  // age
+  if (roles.includes('public')) {
```
marklise, line 426: I think `.includes` needs tweaking.  Public users will only have the `public` role, so the existence of such doesn't necessarily mean they are a public user.  Roles could look like:

`read: ['public', 'sysadmin']`

I think we should do the opposite logic:  Redact unless they include one of `exports.ApplicationRoles` in [misc.js](https://github.com/bcgov/NRPTI/blob/34b82aba6bacea0e1110a216c75b9ab68228b3a5/api/src/utils/constants/misc.js)

dhlevi: Yeah sounds reasonable. I'll make a change.

dhlevi: Updated!

api/src/controllers/search.js
```diff
@@ -419,6 +420,68 @@ let searchCollection = async function(
     }
   });
 
+  // For read only users, we need to redact out the details
+  // of any individual where the birthdate is null or the individual
+  // is less then 19 years old. First step to do this is calculate their
+  // age
+  if (!roles.some(r => constants.ApplicationAdminRoles.indexOf(r) >= 0)) {
+    searchResultAggregation.push({
+      $project: {
```
KitArmstrong, line 429: Nicely done redacting inside the mongo query.

mtCarto (approved):

marklise (approved):

PR #583, opened 2020-09-03
NRPT-387: API and screen for communication package popup

dhlevi:
Addition of a new model called "CommunicationPackage" that contains details for consuming applications to display a popup or other information.

The model contains the following attributes
```
{
    title: string,
    description: string,
    startDate: date,
    endDate: date,
    additionalInfo: object
}
```
Consuming applications can use title, description for the popup, start/end date for when to display. Additional Info is a future-proofing open object that should allow us to include any additional attribute information we need to pass back and forth to the application. Currently not editable by the admin screen, but it should be easy to add an ad-hoc attribute creator for that.

Updates to the config service to add `/api/config/{Application}` get and post endpoints. When calling `/api/config/BCMI` for example, it will append a `COMMUNICATIONS` attribute to the returned package that includes the model. There is only a post included as any changes to the popup will delete the previous and create a new one, which allows us to use the GUID as an indicator for changes.

Added a new tab called Communications to NRPTI admin, and a form that allows you to create these packages for BCMI, LNG and NRCED.

Still requires some cleanup and unit tests, additional debugging. Getting it up here asap so I'm not blocking Kit on his work with the popup on BCMI!

See ticket [NRPT-387](https://bcmines.atlassian.net/browse/NRPT-387)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
+++ b/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
@@ -1,0 +1,113 @@
+import { Component, OnInit, OnDestroy, ViewChild } from '@angular/core';
+import { ActivatedRoute, Router } from '@angular/router';
+import { takeUntil } from 'rxjs/operators';
+import { Subject } from 'rxjs/Subject';
+import { FormGroup, FormControl } from '@angular/forms';
+import { CommunicationsPackage } from '../../../../common/src/app/models/master/common-models/communications-package';
+import { FactoryService } from '../services/factory.service';
+import { DatePickerComponent, LoadingScreenService, Utils } from 'nrpti-angular-components';
+
+@Component({
+  selector: 'communications-add',
+  templateUrl: './communications.component.html',
+  styleUrls: ['./communications.component.scss']
+})
+export class CommunicationsComponent implements OnInit, OnDestroy {
+  @ViewChild(DatePickerComponent) DatePicker: DatePickerComponent;
+
+  private ngUnsubscribe: Subject<boolean> = new Subject<boolean>();
+
+  public myForm: FormGroup;
+
+  public commPackage: CommunicationsPackage;
+  public selectedApplication: string;
+
+  constructor(
+    public route: ActivatedRoute,
+    public router: Router,
+    private factoryService: FactoryService,
+    private utils: Utils,
+    private loadingScreenService: LoadingScreenService
+  ) {}
+
+  ngOnInit() {
+    this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
+      this.commPackage = res.communicationsPackage.COMMUNICATIONS;
+      const urlSegments = this.router.url.split('/');
+      this.selectedApplication = urlSegments[urlSegments.length - 1].toUpperCase();
+
+      this.buildForm();
+    });
+  }
+
+  private buildForm() {
+    this.myForm = new FormGroup({
+      popupTitle: new FormControl({
+        value: (this.commPackage && this.commPackage.title) || '',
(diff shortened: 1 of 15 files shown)


angular/projects/admin-nrpti/src/app/communications/communications.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    public router: Router,
+    private factoryService: FactoryService,
+    private utils: Utils,
+    private loadingScreenService: LoadingScreenService
+  ) {}
+
+  ngOnInit() {
+    this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
+      this.commPackage = res.communicationsPackage.COMMUNICATIONS;
+      const urlSegments = this.router.url.split('/');
```
KitArmstrong, line 36: In the route it is `/communications/:application` so there should be a way to access the `:application` URL param rather than splitting it out here.  I know there is a `paramMap` and a `route.snapshot` that I believe can access it.  This might be overkill though.

angular/projects/admin-nrpti/src/app/communications/communications.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.ngUnsubscribe.next();
+    this.ngUnsubscribe.complete();
+  }
+
+  checkRoles() {
+    if (this.factoryService.userInRole('sysadmin')) {
```
KitArmstrong, line 101: I realize this was not added in this PR, but these user helper functions don't really belong in a factory service, do they?

angular/projects/global/src/lib/services/config.service.ts
```diff
@@ -34,4 +34,16 @@ export class ConfigService {
   get config(): any {
     return this.configuration;
   }
+
+  public getCommunicationPackage(application, pathAPI: string) {
+    return this.httpClient.get(`${pathAPI}/config/${application}`, {});
+  }
+
+  public createCommunicationPackage(communicationPackage, pathAPI: string) {
+    if (!communicationPackage) {
```
KitArmstrong, line 43: a `{}` will pass this check.

KitArmstrong (changes requested):
Everything worked great during testing. Just a few minor comments.

angular/projects/admin-nrpti/src/app/communications/communications.component.ts
dhlevi: Nope, just garbage code put in before doing the routing that I forgot to replace with a proper call to the ActivatedRoute. Updated to `route.snapshot.params.application`

angular/projects/global/src/lib/services/config.service.ts
dhlevi: Added an additional check to ensure `application` exists. The other fields can be null, API will map their lack of existence to null by default, so I don't think we need an overly strong check here.

angular/projects/admin-nrpti/src/app/communications/communications.component.ts
dhlevi: Yeah, I think they're there because factory service had the imports in place for the keycloak service, etc. We can certainly move them... probably a good tech-debt ticket for future enhancement as I don't want to alter all the places where the `isIn<app>Role()` functions are called with this ticket particularly.

KitArmstrong: Yeah, not a worry for this ticket.

KitArmstrong (approved):
👍

danieltruong (approved):

mtCarto (approved):

PR #585, opened 2020-09-04
NRPT-366: Convert ancronyms

dhlevi:
Adding a migration to replace all instances of "Environmental Assessment Office" with EAO and "Ministry of Energy, Mines and Petroleum Resources" with EMPR in the database.

Adding a function in Global Utils to convert the acronyms back to full strings.

Adding functions in NRCED and LNG public to use the global util to display acronyms as the full text (BCMI and NRPTI show acronyms instead).

See ticket [NRPT-366](https://bcmines.atlassian.net/browse/NRPT-366)

### DIFF
--- a/angular/projects/global/src/lib/utils/utils.ts
+++ b/angular/projects/global/src/lib/utils/utils.ts
@@ -105,4 +105,14 @@ export class Utils {
       return fn(...args);
     };
   }
+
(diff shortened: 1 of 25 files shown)


mtCarto (approved):
👍

marklise (approved):

PR #595, opened 2020-09-11
EE-1138 Adds Logging to Apps

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/EE-1138

Replaces console logs in all of the front end apps with the same logging service that is used in BCMI.

### DIFF
--- a/angular/projects/global/src/lib/services/logger.service.ts
+++ b/angular/projects/global/src/lib/services/logger.service.ts
@@ -1,0 +1,75 @@
+import { Injectable, isDevMode } from '@angular/core';
+
+export enum LogLevel {
+  All = 0,
+  Debug = 1,
+  Info = 2,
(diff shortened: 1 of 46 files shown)


angular/projects/global/src/lib/services/logger.service.ts
```diff
@@ -0,0 +1,75 @@
+import { Injectable, isDevMode } from '@angular/core';
```
dhlevi, line 1: Sure looks familiar... but easily one of the most perfect and beautiful pieces of code I've ever seen in my life.

image

dhlevi (approved):
👍 Looks good. I did a quick scan and didn't see any missed console logs that needed to get nuked out.

mtCarto (approved):

PR #598, opened 2020-09-14
NOBUG: Fix Case Sensitivity Issue

KitArmstrong:
Makes the config service case insensitive.


dhlevi (approved):
Looks good. 

Might want to do the same on ln 66 for consistency (though NRPTI is calling the api in uppercase already, so not really an issue):
`newCommPackage.application = args.swagger.params.app.value.toUpperCase();`

danieltruong (approved):

PR #600, opened 2020-09-15
NRPT-315: auto-publish bcmi flavour records

danieltruong:
https://bcmines.atlassian.net/browse/NRPT-315

Collections are automatically public.
Collections require a record to be saved.
Records added to a collection are public.
Records removed from a collection are unpublished.
When creating a record in mines, if not associated with a collection, is not public.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.html
@@ -19,16 +19,6 @@
           <div class="mb-3">
             <span class="grey-subtext">{{ lastEditedSubText }}</span>
           </div>
-          <div>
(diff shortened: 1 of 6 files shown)


marklise (approved):

KitArmstrong (approved):

dhlevi (approved):
Overall looks good. Nice work trimming out a bunch of the now-useless code.

Not sure if it's important for right now, but we can probably remove some of the add/remove role stuff around ln.149
```
if (incomingObj.addRole && incomingObj.addRole === 'public') {
    updateObj['$addToSet'] = { read: 'public' };
```
And just make it always do that, so anyone who hits API direct will still save collections as always-published. And just remove the removeRole stuff as you can't actually remove the role by business rules. Again, it's only to prevent direct hits to the API as the UI handles it so a pretty low risk thing.

MaxWardle (approved):

PR #602, opened 2020-09-16
NRPT-315: Fixed typo

danieltruong:


angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
```diff
@@ -310,6 +310,8 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
       }
     });
 
+    console.log(this.myForm.get('collectionRecords')['controls']);
```
mtCarto, line 313: is there a logger in this component we can use instead?

mtCarto (approved):

PR #606, opened 2020-09-17
NOBUG: Fixes Core Import Collections

KitArmstrong:
This updates the Core importer to work with the changes that were made to the collections controller.

After some discussion the function name was updated as it no longer only creates records.


danieltruong (approved):

MaxWardle (approved):
looks good :+1:

PR #622, opened 2020-09-22
NRPT-424 Fix Management Plan Record type not saving agency

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-424

Management Plan record type was saving agency under the agency field.  Updated it to save agency under issuingAgency like other record types.


KitArmstrong (changes requested):
We will need to create a migration to update any existing Management Plan records with the new `issuingAgency` field while also remove the `agency` field.

PR #624, opened 2020-09-24
NRPT-424 Fix Records agency related issues

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-424

The original goal of this PR was to fix Management Plan records not saving agency correctly.  The work has been expanded to standardize records agency fields to issuingAgency.

This PR includes 4 commits:

1. Update all record types in frontend to use issuingAgency instead of agency

2. Update all record types in API to use issuingAgency instead of agency

3. Update EPIC importer to set default agency as EAO instead of Environmental Assessment Office

4. Add migration script to fix existing records agency issues


api/migrations/20200924004015-convertAgencyToIssuingAgency.js
```diff
@@ -0,0 +1,90 @@
...
+exports.down = function(db) {
+  return null;
+};
+
+exports._meta = {
+  "version": 1
+};
+
+async function renameAgency(nrpti) {
+  const schemas = ["ManagementPlan", "ManagementPlanLNG"];
```
KitArmstrong, line 53: Looks like "ConstructionPlan" and "ConstructionPlanLNG" also had its property changed.

KitArmstrong (changes requested):
Just the one comment about adding the two additional record types to the migration.

api/migrations/20200924004015-convertAgencyToIssuingAgency.js
BcGovNeal: I didn't find any issues with the existing data for ConstructionPlan and ConstructionPlanLNG so they weren't included in that function.

KitArmstrong: Below it looks like the models were changed though.  Did they just not have the 'agency' property on the ones in the DB some how?

Sorry, I could be missing something.

BcGovNeal: Yeah the model had "agency", but the DB had "issuingAgency".  So ConstructionPlan and ConstructionPlanLNG didn't need to be fixed in the DB.

KitArmstrong (approved):

marklise (approved):

PR #625, opened 2020-09-24
NRPT-432: remove isPublished flags from flavour records

mtCarto:
* [NRPT-432](https://bcmines.atlassian.net/browse/NRPT-432)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
+++ b/angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
@@ -414,7 +414,7 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
             alert('Failed to create collection.');
           } else {
             this.loadingScreenService.setLoadingState(false, 'main');
-            this.router.navigate(['mines', this.collection._master, 'collections', this.collection._id, 'detail']);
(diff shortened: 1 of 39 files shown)


angular/projects/admin-nrpti/src/app/mines/mines-collections-add-edit/mines-collections-add-edit.component.ts
```diff
@@ -414,7 +414,7 @@ export class MinesCollectionsAddEditComponent implements OnInit, OnDestroy {
             alert('Failed to create collection.');
           } else {
             this.loadingScreenService.setLoadingState(false, 'main');
-            this.router.navigate(['mines', this.collection._master, 'collections', this.collection._id, 'detail']);
+            this.router.navigate(['mines', this.collection.project, 'collections', this.collection._id, 'detail']);
```
KitArmstrong, line 417: Nice catch!

KitArmstrong (changes requested):
With a quick search it looks like 'isBcmiPublished' is being checked in at least one component's HTML. As per the last AC on this ticket it that flag should only ever be checked for filtering. HTML should always look at the read array.

mtCarto:
Cleaned up other references to isBcmiPublished flag in table-row components. Should be all of them now

KitArmstrong (approved):

MaxWardle (approved):

PR #631, opened 2020-09-30
NRPT-309: right aligned all records dropdown

MaxWardle:
https://bcmines.atlassian.net/browse/NRPT-309


mtCarto (approved):

KitArmstrong (approved):
Worked in all three browsers I tested with.

PR #632, opened 2020-10-01
NRPT-454: Removes CollectionID From Master Records

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-454

Changes: Removes the "collectionId" fields from all of the master records. Functionality should remain the same as it did before.  BCMI flavours will hold the reference to the collection.


danieltruong:
This is the ticket thats blocking my flavour records ticket that I've been working on. For some reason it wasn't assigned to me

This ticket also would require a migration script to remove the collectionId field from all master existing master records. We also need a revision on the mines records lists because they all rely on collectionId to do a populate. Both of these issues are addressed in the PR I am currently working on. It also includes the model change you have in this PR.

PR #633, opened 2020-10-01
NRPT-429 and NRPT-454: BCMI collection revisions

danieltruong:
Collection uses flavour id instead of master. Removing collectionId from master records


KitArmstrong (changes requested):
Appears to be an issue with created records.  After I associate a new record to a collection it no longer appears in the Mine Record List.  Even if I remove it from the collection it does not return.

Steps to reproduce:
- Go to a Mine's Record List page
- Create a new record 
- Select the new record and choose "Add to Existing Collection"
- Update the collection
- Return to the Mine's Record List page and search for the record that was created

This same issue occurs if the mine is created from inside a collection too.

danieltruong:
> Appears to be an issue with created records. After I associate a new record to a collection it no longer appears in the Mine Record List. Even if I remove it from the collection it does not return.
> 
> Steps to reproduce:
> 
> * Go to a Mine's Record List page
> * Create a new record
> * Select the new record and choose "Add to Existing Collection"
> * Update the collection
> * Return to the Mine's Record List page and search for the record that was created
> 
> This same issue occurs if the mine is created from inside a collection too.

I can't seem to recreate this.

danieltruong:
> Appears to be an issue with created records. After I associate a new record to a collection it no longer appears in the Mine Record List. Even if I remove it from the collection it does not return.
> 
> Steps to reproduce:
> 
> * Go to a Mine's Record List page
> * Create a new record
> * Select the new record and choose "Add to Existing Collection"
> * Update the collection
> * Return to the Mine's Record List page and search for the record that was created
> 
> This same issue occurs if the mine is created from inside a collection too.

48cfb676f269f621135169c8388ec0fc55ee04b7

marklise (approved):

KitArmstrong (approved):

PR #635, opened 2020-10-05
NRPT-438: return collection docs in sort order defined by collection

mtCarto:
Sort order is defined by a collection's records array, the mongo lookups did not preserve order. Docs should now be returned in the proper order.

- [NRPT-438](https://bcmines.atlassian.net/browse/NRPT-438)

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -897,17 +897,40 @@ const executeQuery = async function (args, res, next) {
       {
         $match: { _id: mongoose.Types.ObjectId(args.swagger.params._id.value) }
       },
-      // lookup the records
+      // lookup the records and set order as defined by collection records array
       {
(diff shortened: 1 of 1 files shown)


KitArmstrong (approved):

BcGovNeal:
I tried running this locally and the queries seem to be very slow on the BCMI public searches.  It's causing the data tabs on the mines to not load correctly.

I tried this with both my local mongodb and mongodb in OCP test.  I think this should be looked at before merge.

mtCarto:
I'll take a look, I was concerned that there might be a slow down.

BcGovNeal (approved):
Works much better

mtCarto:
Yeah, turns out those pipeline operations in a lookup are super slow

PR #641, opened 2020-10-15
NRPT-480: Cleaning up issues with epic importer. Adding missing models

danieltruong:


mtCarto (approved):

mtCarto (approved):

api/migrations/20200709192021-mineMapTable.js
```diff
@@ -134,7 +134,7 @@ const mineMaps = [
   * We receive the dbmigrate dependency from dbmigrate initially.
   * This enables us to not have to rely on NODE_PATH.
```
KitArmstrong, line 135: Can we confirm this has never been run in prod? It is not one of the migrations we run as part of the data import process so I am slightly concerned it might get missed.

KitArmstrong (approved):

api/migrations/20200709192021-mineMapTable.js
danieltruong: ef06a1dc5839d3e8fb6860dd37f1de3b4d365a0d

PR #642, opened 2020-10-16
NPRT-289: Automates BCOGC Import

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-289

Updates the BCOGC importer to do an automatic import rather than manual CSV upload. Also a migration as the source system was changed so that it makes more sense in the filters.


BcGovNeal:
Does this need a cron job?

BcGovNeal (approved):

MaxWardle (approved):
Importer seems to run well

PR #659, opened 2020-10-26
NRPT-503: Collection Updates

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-503

Part two of this ticket's fixes. Updated the collection controller to set new collection's published status based on that of the mine. Also updated to set the document's status based on that of the collection (which is based on the mine).

### DIFF
--- a/api/src/controllers/collection-controller.js
+++ b/api/src/controllers/collection-controller.js
@@ -8,7 +8,7 @@ const { userHasValidRoles } = require('../utils/auth-utils');
 const utils = require('../utils/constants/misc');
 const RECORD_TYPE = require('../utils/constants/record-type-enum');
 const mongoose = require('mongoose');
-const ObjectID = require('mongodb').ObjectID;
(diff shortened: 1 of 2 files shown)


mtCarto (approved):
Codel looks good, records show as unpublished after a mine is publishend then unpublished.

api/src/controllers/collection-controller.js
```diff
@@ -368,29 +382,38 @@ const checkRecordExistsInCollection = async function (records, collectionId, edi
...
+      permissionAction = {
+        $addToSet: {
+          read: 'public'
         }
-      ));
+      };
     }
+    // Ensure the record has the collectionId set
+    // Also, if we are associating a record
+    promises.push(collectionDB.findOneAndUpdate(
```
BcGovNeal, line 408: Could just use updateOne if only updating and not returning data

BcGovNeal (approved):

PR #668, opened 2020-10-28
NRPTI-496-2: refine the collection lookup query and fix existing collections

mtCarto:
[NRPT-496](https://bcmines.atlassian.net/browse/NRPT-496)
 - removes records from existing collections that they don't belong to

### DIFF
--- a/api/migrations/20200709183253-loadMemDocs.js
+++ b/api/migrations/20200709183253-loadMemDocs.js
@@ -124,6 +124,7 @@ exports.up = async function (db) {
       }
       if (nrptiMine) {
         // Found a mine, now lets create a record and upload up the docs
+        console.log(`Current mine is ${nrptiMine.name},   id: ${nrptiMine._id}`);
         for (const collection of mine.collectionData) {
           console.log(`Processing collection ${collection.displayName}`);
           if (!collection.displayName) {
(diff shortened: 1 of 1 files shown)


api/migrations/20200709183253-loadMemDocs.js
```diff
@@ -213,22 +216,36 @@ exports.up = async function (db) {
...
-                  console.log(`Adding doc to docsArray ${existingDoc._id}`);
+                  console.log(`Adding doc ${existingDoc._id} to docsArray of ${bcmiCollection._id}`);
                   allNewDocs.push(existingDoc._id);
                 }
               }
               docsExisting += 1;
+              // Ensure that records only exist in the proper collection for their mine
+              const collectionsToFix = await nrpti.find({records: { $in: [ existingDoc._id ]}});
+              collectionsToFix.forEach( async (coll) => {
+                if (!coll.project.equals(nrptiMine._id)) {
```
KitArmstrong, line 234: Does this work. `.equals` is Java if I remember correctly. 
`(coll.project !== nrptiMine._id)`

mtCarto: It should be an mongo objectid method, but I don't explicitly cast to obj id. Might just be strings.. It's also used for the existing collection id comparison on ln 218. I'll double check this

mtCarto: This should be correct the, the strict comparison operator doesn't seem to work on ObjectIds. Despite being console logged as a string I think they are using the mongo object id in the script.

KitArmstrong (approved):

BcGovNeal (approved):

PR #670, opened 2020-10-28
NOBUG: Edit mines does not pub/unpub collections, records and docs

danieltruong:


api/src/controllers/put/mine-bcmi.js
```diff
@@ -64,10 +67,12 @@ exports.editMaster = function(args, res, next, incomingObj) {
     updateObj.$addToSet['read'] = 'public';
     updateObj.$set['datePublished'] = new Date();
     updateObj.$set['publishedBy'] = args.swagger.params.auth_payload.displayName;
+    await collectionController.publishCollections(ObjectId(mineId), args.swagger.params.auth_payload);
```
KitArmstrong, line 70: I think the casting to OID should happen in the `publishCollections` method so it can be used anywhere without worrying about calling it with casted params.

danieltruong: a8fe229

BcGovNeal (approved):

mtCarto (approved):

PR #681, opened 2020-11-05
NRPT-533: Fixes Issues With EPIC Records

KitArmstrong:
Ticket: https://bcmines.atlassian.net/browse/NRPT-533

Records from EPIC where missing for these two mines; however, the records were available in the All Records page.  These mines were not mapped and therefore the records were not added correctly. This migration links the orphaned records to the correct mines.

### DIFF
--- a/api/migrations/202011051807132-addMissingMineGuid.js
+++ b/api/migrations/202011051807132-addMissingMineGuid.js
@@ -33,7 +33,7 @@ exports.up = async function(db) {
         'CertificateAmendmentBCMI',
         'ManagementPlanBCMI'
       ]},
-      mineGuid: null
+      mineGuid: { $in: [null, ''] }
     },
     {  $set: { mineGuid: '739e90ee-807c-4287-91a3-7099352e1ffa' } });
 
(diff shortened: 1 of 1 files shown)


api/migrations/202011051807132-addMissingMineGuid.js
```diff
@@ -0,0 +1,69 @@
...
+    await nrpti.updateMany({ 
+      _epicProjectId: ObjectId('58851156aaecd9001b81e652'),
+      _schemaName: { $in: [
+        'CertificateBCMI',
+        'OrderBCMI',
+        'InspectionBCMI',
+        'CertificateAmendmentBCMI',
+        'ManagementPlanBCMI'
+      ]},
+      mineGuid: null
```
BcGovNeal, line 51: Do records where `mineGuid: ""` also require fixing?

KitArmstrong: I could add it as a safety, but the particular records we are targeting will have it set to `null` and there are none with a blank string.

BcGovNeal: Ah I see the data on the live db has been altered.  I do have a bunch of records in my copy of the post data load DB with just empty string mineGuid.  Might still be a good idea to add blank string check.

BcGovNeal (approved):

api/migrations/202011051807132-addMissingMineGuid.js
KitArmstrong: Updated to check for a bank string too.

mtCarto (approved):

PR #684, opened 2020-11-10
NRPT-538 Fix  Core and Core-doc importers not bringing in new data

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-538

Core importer was checking for new permits on existing mines based on the number of permits in the database.  This does not work if there are additional sources of permits such as mem-admin.  The permit count check has been removed and now the importer just simply compares `permit_amendment_document_guid` on the incoming data for new permits.

Also fixed Core-doc importer not reporting number of records processed.

### DIFF
--- a/api/src/integrations/core-documents/datasource.js
+++ b/api/src/integrations/core-documents/datasource.js
@@ -44,6 +44,7 @@ class CoreDocumentsDataSource {
 
       // Run main process.
       await this.updateRecords();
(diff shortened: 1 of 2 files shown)


mtCarto (approved):
Tested out locally, no issues found. Code looks good

KitArmstrong (approved):

PR #685, opened 2020-11-10
NRPT-537: fix permittee and permit number not updating when core data does

mtCarto:
[NRPT-537](https://bcmines.atlassian.net/browse/)

### DIFF
--- a/api/src/integrations/core/datasource.js
+++ b/api/src/integrations/core/datasource.js
@@ -216,6 +216,7 @@ class CoreDataSource {
         if (permitInfo.permit.permit_amendments && permitInfo.permit.permit_amendments.length) {
           await this.createCollections(collectionUtils, permitInfo.permit, savedRecord[0].object[0]);
         }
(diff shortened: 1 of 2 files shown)


api/src/integrations/core/datasource.js
```diff
@@ -384,6 +385,14 @@ class CoreDataSource {
       const promises = preparedPermits.map(permit => permitUtils.createItem(permit));
       await Promise.all(promises);
     }
+
+    // get permit info from transformed core record to ensure permittee + permit num are updated
+    const permitToUpdate = await this.getMinePermit(nrptiRecord);
```
BcGovNeal, line 390: I don't think this line is needed since it should have the same value as the existingPermit on line 351.  So `nrptiRecord` isn't also needed in this function

mtCarto: Oh that's right, removed this extra lookup.

mtCarto:
Waiting for [NRPT-538](https://github.com/bcgov/NRPTI/pull/684).

KitArmstrong (approved):

BcGovNeal (approved):

PR #690, opened 2020-11-18
NRPT-549 Fix Core importer not updaing collections and records

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-549

- Added `20201117185607-addCoreCollectionRefId.js` migration to associate collections with the `permit_amendment_guid` from Core using the existing `_sourceRefId` on permit records
- Refactored Collection controller so the update function can be called directly from another file
- Added logic to update collections and records if changes are detected

Make sure to run the migration first if testing locally

### DIFF
--- a/api/src/controllers/collection-controller.js
+++ b/api/src/controllers/collection-controller.js
@@ -44,10 +44,7 @@ exports.protectedPut = async function (args, res, next) {
   // Confirm user has correct role for this type of record.
   if (!userHasValidRoles([utils.ApplicationRoles.ADMIN, utils.ApplicationRoles.ADMIN_BCMI], args.swagger.params.auth_payload.realm_access.roles)) {
     throw new Error('Missing valid user role.');
-  }
-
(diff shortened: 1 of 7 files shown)


api/src/controllers/collection-controller.js
```diff
@@ -64,8 +61,26 @@ exports.protectedPut = async function (args, res, next) {
     defaultLog.info(`protectedPut - you must provide an object`);
     queryActions.sendResponse(res, 400, {});
     next();
+  }  
+
+  let obj = null;
+  try {
+    // obj = await Put.updateById(collectionId, updateObj);
```
mtCarto, line 68: remove this commented line

mtCarto (approved):
Collection is updated appropriately 👍

api/src/controllers/collection-controller.js
BcGovNeal: Fixed in https://github.com/bcgov/NRPTI/pull/690/commits/2bd4d09487e3927447fa6e30c778887452e5a60b

MaxWardle (approved):
Migration runs fine

PR #701, opened 2020-11-30
NRPT-577: add objectId is valid check on id param

mtCarto:
[NRPT-577](https://bcmines.atlassian.net/browse/NRPT-577)

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -799,6 +799,15 @@ const executeQuery = async function (args, res, next) {
   if (dataset[0] === 'Item') {
     defaultLog.info('ITEM GET', { _id: args.swagger.params._id.value });
 
+    if (!args.swagger.params._id.value || !ObjectID.isValid(args.swagger.params._id.value)) {
+      defaultLog.info(`Error searching for item: ${args.swagger.params._id.value}, Error: Invalid Item ID supplied`)
+      return QueryActions.sendResponse(
+        res,
+        400,
+        `Error searching for item ${args.swagger.params._id.value}. Invalid item id supplied`
+      );
+    }
+
     let aggregation = [
       {
         $match: { _id: mongoose.Types.ObjectId(args.swagger.params._id.value) }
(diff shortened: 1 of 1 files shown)


MaxWardle (commented):
seems like it might be blocking all item gets now:
`
info: ITEM GET _id=5f80a3b37f0b300021abd018
`
`
info: Error searching for item: 5f80a3b37f0b300021abd018, Error: Invalid Item ID supplied
`

api/src/controllers/search.js
```diff
@@ -798,6 +798,17 @@ const executeQuery = async function (args, res, next) {
 
   if (dataset[0] === 'Item') {
     defaultLog.info('ITEM GET', { _id: args.swagger.params._id.value });
+    // It's possible to bring down the API if you pass in an invalid item ID, or null.
+    // a post to "api/search?dataset=Item&_id=some-non-object-id" will cause the server to die.
+    // This will still let any 12 char value through (it passes isValid)
+    if (!args.swagger.params._id || !ObjectID.isValid(args.swagger.params._id)) {
```
BcGovNeal, line 804: Should be `args.swagger.params._id.value` instead of `args.swagger.params._id` on this line.

BcGovNeal (changes requested):

api/src/controllers/search.js
mtCarto: good catch 👍

BcGovNeal (approved):

api/src/controllers/search.js
```diff
@@ -738,6 +738,17 @@ const generateMatchesForAggregation = async function (and, or, nor, searchProper
 };
 
 const executeQuery = async function (args, res, next) {
+    // It's possible to bring down the API if you pass in an invalid item ID, or null.
+    // a post to "api/search?dataset=Item&_id=some-non-object-id" will cause the server to die.
+    // This will still let any 12 char value through (it passes isValid)
```
marklise: To protect against this, the final call should be wrapped in a try/catch and handled gracefully, instead of allowing the aggregation to crash the API.

marklise (changes requested):

api/src/controllers/search.js
mtCarto: Ahh I was mistaken here, a 12 char value will not crash the api.  It will return no results. I will implement the try catch, to prevent other errors crashing it out.

marklise (approved):

PR #702, opened 2020-12-01
NRPT-365: import bcogc warnings from csv table

mtCarto:
[NRPT-365](https://bcmines.atlassian.net/browse/NRPT-365)

### DIFF
--- a/api/src/models/lng/warning-lng.js
+++ b/api/src/models/lng/warning-lng.js
@@ -6,6 +6,7 @@ module.exports = require('../../utils/model-schema-generator')(
     _schemaName: { type: String, default: 'WarningLNG' },
     _epicProjectId: { type: 'ObjectId', default: null, index: true },
     _sourceRefId: { type: 'ObjectId', default: null, index: true },
+    _sourceRefOgcWarningId: { type: String, default: null },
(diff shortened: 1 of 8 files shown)


api/src/models/lng/warning-lng.js
```diff
@@ -6,6 +6,7 @@ module.exports = require('../../utils/model-schema-generator')(
     _schemaName: { type: String, default: 'WarningLNG' },
     _epicProjectId: { type: 'ObjectId', default: null, index: true },
     _sourceRefId: { type: 'ObjectId', default: null, index: true },
+    _sourceRefOgcWarningId: { type: String, default: null },
```
danieltruong, line 9: Is there a reason why we have another field for this? Is it possible to use _sourceRefId?

BcGovNeal: `_sourceRefId` is `ObjectId` and the ref id here is a `string`

mtCarto: yeah the incoming field used as a reference is a string `Title`

danieltruong (approved):

BcGovNeal (approved):

PR #707, opened 2020-12-08
NRPT-582 Add public read to master records with published flavours

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-582

### DIFF
--- a/api/migrations/20201208212602-addPublicToMaster.js
+++ b/api/migrations/20201208212602-addPublicToMaster.js
@@ -1,0 +1,81 @@
+'use strict';
+
+const { MASTER_SCHEMA_NAMES } = require('../src/utils/constants/misc');
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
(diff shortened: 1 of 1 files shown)


mtCarto (approved):
Migration looks good, verified that `wildfire` role is able to see published records now.

api/migrations/20201208212602-addPublicToMaster.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const { MASTER_SCHEMA_NAMES } = require('../src/utils/constants/misc');
+
```
marklise, line 4: Given this is a migration to be run at a specific time and not in the future if those misc constants change, I think we should be more explicit here instead of referencing.

BcGovNeal: Fixed in https://github.com/bcgov/NRPTI/pull/707/commits/57bddd711f91628562e2f6bb92e7153c7660f420

danieltruong (approved):

PR #709, opened 2020-12-08
NRPT-518: unpublishing Order other records for NRCED

danieltruong:


marklise (approved):

api/src/integrations/epic/base-record-utils.js
```diff
@@ -242,10 +242,22 @@ class BaseRecordUtils {
...
-        createObj[this.recordType.flavours.nrced._schemaName] = {
-          summary: nrptiRecord.description || '',
-          addRole: 'public'
-        };
+        // Only EPIC records with Milestone = "Compliance and Enforcement" and
+        // Document Type = "Inspection Record" or "Order" should be auto-published to NRCED.
+        if (
+          (nrptiRecord._schemaName === 'Order' && nrptiRecord._epicMilestoneId == '5cf00c03a266b7e1877504ef') ||
+          (nrptiRecord._schemaName === 'Order' && nrptiRecord._epicMilestoneId == '5df79dd77b5abbf7da6f5201') ||
+          (nrptiRecord._schemaName === 'Inspection')
```
BcGovNeal, line 250: Do all `Inspection` records have `Compliance and Enforcement` Milestone?  If not, then doesn't this line publish all `Inspection` regardless of the Milestone?

mtCarto: It looks like `Inspection` records in the system now only have C&E milestones, but not sure we can guarantee that will always be the case with data coming from EPIC.  Probably best to include the milestoneId as a safeguard

mtCarto (commented):
Verified the migration locally, apart from the comment around the `Inspection` & milestones it looks good.

api/src/integrations/epic/base-record-utils.js
marklise: I believe it is correct to say: every `Inspection` will be `C&E` given the nature of what they are doing.

mtCarto: 👍

mtCarto (approved):

api/src/integrations/epic/base-record-utils.js
danieltruong: 5b69fb7

marklise (approved):

BcGovNeal (approved):

PR #710, opened 2020-12-08
NRPT-556: update search controller to redact issuedTo based on role

mtCarto:
- [NRPT-556](https://bcmines.atlassian.net/browse/NRPT-556)

### DIFF
--- a/api/src/controllers/search.js
+++ b/api/src/controllers/search.js
@@ -7,7 +7,8 @@ let qs = require('qs');
 let mongodb = require('../utils/mongodb');
 let moment = require('moment');
 let fuzzySearch = require('../utils/fuzzySearch');
-let constants = require('../utils/constants/misc')
+let constants = require('../utils/constants/misc');
+const AppRoles = constants.ApplicationRoles;
 
 function isEmpty(obj) {
   for (const key in obj) {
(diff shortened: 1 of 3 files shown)


api/src/controllers/search.js
```diff
@@ -7,7 +7,8 @@ let qs = require('qs');
 let mongodb = require('../utils/mongodb');
 let moment = require('moment');
 let fuzzySearch = require('../utils/fuzzySearch');
-let constants = require('../utils/constants/misc')
+let constants = require('../utils/constants/misc');
+let AppRoles = constants.ApplicationRoles;
```
marklise, line 11: Unless modifying, technically should be `const`

BcGovNeal (approved):

BcGovNeal (changes requested):
I test this again and it seems the `issuedTo` names are showing on the Record list when `issuedTo.read` doesn't have `admin:wf` but `read` does.  Record detail is redacting correctly though.

mtCarto:
I'll take a look and correct this

mtCarto:
Should be working as expected now, requires the extra migration to prevent redaction of company names.

BcGovNeal (approved):
IssuedTo is displaying as expected

MaxWardle (approved):
Very nice 👏

PR #711, opened 2020-12-09
NRPT-559: added wf role to record add edit, pre unit tests

MaxWardle:
Creating this for visibility while I continue work on unit tests. Functionality should be there


MaxWardle:
duplicated changes

PR #712, opened 2020-12-10
NRPT-520: Making dataset a manditory field when searching

danieltruong:


api/src/controllers/search.js
```diff
@@ -738,9 +738,13 @@ const generateMatchesForAggregation = async function (and, or, nor, searchProper
 };
 
 const executeQuery = async function (args, res, next) {
+  if (!args.swagger.params.dataset.value) {
+    defaultLog.info('Bad Request');
+    QueryActions.sendResponse(res, 400, {});
```
BcGovNeal, line 743: Execution needs to stop here.  The search API still crashes if the code execution continues.

BcGovNeal (changes requested):

api/src/controllers/search.js
danieltruong: 997c247

BcGovNeal (approved):
Looks good!

MaxWardle (approved):

PR #719, opened 2020-12-15
NRPT-392 Fix master record isNrcedPublished and isLngPublished flags

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-392

This PR cycles through all master records and sets their isNrcedPublished and isLngPublished flags properly.  The original ticket is only for records from LNG-CSV and NRIS-EPD, but I see no reason not to fix all records.

### DIFF
--- a/api/migrations/20201215225546-fixFlavourPublishedStatus.js
+++ b/api/migrations/20201215225546-fixFlavourPublishedStatus.js
@@ -1,0 +1,109 @@
+'use strict';
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {};
+
+/**
+ * This migration sets the isLngPublished and isNrcedPublished flags on master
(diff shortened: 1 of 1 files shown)


mtCarto (approved):

danieltruong (changes requested):
At the moment, I don't think the EPIC import sets isNrcedPublished and isLngPublished. This should probably be updated along with any other imports that might not have these flags set.

BcGovNeal:
> At the moment, I don't think the EPIC import sets isNrcedPublished and isLngPublished. This should probably be updated along with any other imports that might not have these flags set.

@danieltruong The importers all use the Record Controller to create records and the Record Controller already properly sets the "isPublished" fields for flavours.  https://github.com/bcgov/NRPTI/blob/6b8ef5f45cf38cb25d442b3df4f73efc34cc6da0/api/src/utils/post-utils.js#L77-L81

The ticket was originally meant to fix certain `lng-csv` and `nris-epd`, but I made to migration to just simply target every record type instead.

danieltruong (approved):

PR #727, opened 2020-12-17
NRPT-392 Fix master flavour isPublished status and _flavourRecords asscociations

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-392

Updating this migration to include master _flavourRecords fixes.  There are a number of records created by the NRIS and EPIC importer between July and September 2020 that didn't have the master records' _flavourRecords array set correctly.  The root cause seems to be resolved, this just fixes existing data.

**Note:** Make sure to delete `20201215225546-fixFlavourPublishedStatus` record from migration table after merging


mtCarto (approved):
Looks good, local test cleaned up those missing flavours

marklise (approved):

PR #743, opened 2021-01-08
NRPT-607: changed EMPR to EMLI in multiple places

MaxWardle:
[NRPT-607](https://bcmines.atlassian.net/browse/NRPT-607): migration to change all instances of EMPR on records and collections to EMLI, as well as some front end data manipulation to show the full text in the dropdowns.


MaxWardle:
Missed a spot

PR #748, opened 2021-01-15
NRPT-569: Env yamls for OCP4

marklise:


BcGovNeal (changes requested):
A few things from my point of view:

- Please convert these `yaml` files to [OpenShift templates](https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-templates).  They should be compatible with `oc process -f file.yaml` command
- Please remove the OpenShift managed fields like `selfLink`, `resourceVersion`, `manageFields` and others.  Basically if it's not something you specified then it should be removed from the templates.
- Routes and services configs should be combined with the deployment config yaml of the same component
- The image stream configs should be combined with the build configs.  I can handle these as part of the Jenkins ticket
- We already have build and deploy configs in the `tools` directory.  We should update those instead of creating new ones elsewhere.   Also the frontend `yaml` files probably don't belong in the `api` directory

PR #751, opened 2021-01-20
NRPT-526: add AGRI MIS inspections csv importer

mtCarto:
[NRPT-526](https://bcmines.atlassian.net/browse/NRPT-526)

### DIFF
--- a/angular/projects/common/src/app/utils/record-constants.ts
+++ b/angular/projects/common/src/app/utils/record-constants.ts
@@ -223,6 +223,7 @@ export class Picklists {
   };
 
   public static readonly sourceSystemRefPicklist = [
+    'agri-mis',
     'bcogc',
     'core',
     'coors-csv',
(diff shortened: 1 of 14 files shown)


angular/projects/common/src/app/utils/record-constants.ts
```diff
@@ -232,7 +232,8 @@ export class Picklists {
     'nris-epd',
     'nro-inspections-csv',
     'nrpti',
-    'ocers-csv'
+    'ocers-csv',
+    'agri-mis'
```
BcGovNeal, line 236: nitpick, put `agri-mis` at the top so the list is alphabetically sorted on the UI. Otherwise it looks great!

BcGovNeal (approved):

angular/projects/common/src/app/utils/record-constants.ts
mtCarto: 👍

marklise (approved):

PR #753, opened 2021-01-21
NRPT-638: added redacted record materialized view forpublic searching

MaxWardle:
[NRPT-638](https://bcmines.atlassian.net/browse/NRPT-638):  

NOTE: this PR goes beyond the scope of this ticket and will affect all public searches. All public searches will now use the redacted record subset to avoid data leaks.

To test this you can grab an access token and run this `curl -H "Authorization: Bearer ${TOKEN}" -H "Content-Type: application/json" -d '{"taskType":"updateMaterializedView", "materializedViewSubset":"redactedRecordSubset"}' http://localhost:3000/api/task`

then run the migration and the bug described in the ticket should be gone, but the rest of search on nrced public should be working.


mtCarto (approved):

mtCarto:
Looks good, the record does not show when searching for the issued to subject's name. Other public searches seem to be working as expected

BcGovNeal (approved):

PR #755, opened 2021-01-21
NRPT-80: add csv import for  coors court convictions

mtCarto:
[NRPT-80](https://bcmines.atlassian.net/browse/NRPT-80)


mtCarto:
Updated the importer to handle records spanning multiple csv rows and ensure penalties are appended to existing records where possible. 
Note: it may look like there are duplicates in the records list, but it should be the case that each of those 'duplicates' has a different piece of legislation associated with it.

BcGovNeal (approved):

danieltruong (approved):

PR #757, opened 2021-01-25
NRPTI-597 Migrate NRPTI backup containers

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-597

Note:  I added the Postgres backup configs as well even though it's not currently used.  It can be used when NRPTI moves to Postgres for Metabase


MaxWardle (approved):
Looks good, nice simplifications

tools/backups/db-backup/mongodb-backup.dc.yaml
```diff
@@ -0,0 +1,185 @@
...
+                  cpu: 250m
+                  memory: 500Mi
+                requests:
+                  cpu: 10m
+                  memory: 50Mi
+              volumeMounts:
+                - name: nrpti-${NAME}
+                  mountPath: /backups/
+                - name: nrpti-${NAME}-verification
+                  mountPath: /var/lib/pgsql/data
```
mtCarto, line 144: Can't recall if the backup scripts care about the specific paths for the db's. Current DC has this set as `/var/lib/mongodb/data`

BcGovNeal: Yes you're correct!  It should be `/var/lib/mongodb/data`.  Fixed in https://github.com/bcgov/NRPTI/pull/757/commits/09cee3a4cf7c244a41ce53edeb00a800b02594a6

marklise (approved):

PR #759, opened 2021-01-25
Update DC for OCP4

marklise:


danieltruong (approved):

tools/metabase-dc.json
```diff
@@ -1,337 +1,340 @@
...
+        ],
+        "resources": {
+          "requests": {
+            "storage": "${METABASE_PVC_SIZE}"
+          }
+        }
+      }
+    },
+    {
+      "kind": "ImageStream",
```
BcGovNeal, line 95: This `ImageStream` section isn't needed.  All this does is create an empty `ImageStream` object in the prod namespace.

marklise: Fixed in #af72ccd

BcGovNeal (approved):

PR #764, opened 2021-02-02
NRPT-646: Fix bug in import screen load, sanity check on missing issu…

marklise:
…edTo props, update remote endpoint, change to await on each proimise for slower consumption of resource.

### DIFF
--- a/angular/projects/admin-nrpti/src/app/services/import.service.ts
+++ b/angular/projects/admin-nrpti/src/app/services/import.service.ts
@@ -98,7 +98,7 @@ export class ImportService {
           )
         );
       }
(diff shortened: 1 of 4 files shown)


mtCarto (approved):

BcGovNeal:
Are the records with missing `issuedTo` getting fixed in another PR?

marklise:
> Are the records with missing `issuedTo` getting fixed in another PR?

Not every record type gets an 'issuedTo' - ie certificate

BcGovNeal:
> > Are the records with missing `issuedTo` getting fixed in another PR?
> 
> Not every record type gets an 'issuedTo' - ie certificate

Got it

BcGovNeal (approved):

PR #775, opened 2021-02-12
NRPT-611: Added AGRI-CMDB CSV Importer

MaxWardle:
[NRPT-611](https://bcmines.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=NRPT-611)

To test this use the csv found in the ticket and check that it creates 6 records.


api/src/importers/cmdb/datasource.js
```diff
@@ -0,0 +1,162 @@
...
+
+      // construct the outcome description by appending the regulation sections of rows with the same inspection id
+      for (let i = 0; i < this.csvRows.length; i++) {
+        let inspectionId = this.csvRows[i]['inspection id'];
+        let regulationSection = this.csvRows[i]['regulation section'];
+        let outcomeDescription = 'Compliance issue(s) identified under the following acts or regulations: ' + regulationSection;
+
+        if (recordInspectionIds.includes(inspectionId)) {
+          let index = recordInspectionIds.indexOf(inspectionId);
+          outcomeDescription = recordOutcomeDescriptions[index] + '; ' + regulationSection;
```
mtCarto, line 68: Duplicate checking for regulationSection would be good to have here. Several of the records have multiple copies of the same regulationSection value.
eg. Company 7 outcomeDescription has 2 FSLR-s.32 values

mtCarto (changes requested):

api/src/importers/cmdb/datasource.js
MaxWardle: good catch thanks, updated

mtCarto (approved):
Looks good, duplicate checking + updates are working as expected. Just need unit tests passing

BcGovNeal (approved):

PR #781, opened 2021-02-19
NRPT-11 ALC Inspections CSV import

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-11

This PR adds the ALC Inspection CSV import.  Test file is in Jira ticket

Note:  I decided to use a field called `_sourceRefStringId` instead of something like `_sourceRefAlcInspectionId` for tracking source system IDs.  This is because as we add more import sources, the number of these unique `_sourceRef<system>Id` is starting to get out of hand.  The purpose of `_sourceRefStringId` field is have it be used across all importers instead of creating an unique field for each importer.

The reason `_sourceRefId` isn't used for this is because it is of type `ObjectId` in the models and won't store other types

### DIFF
--- a/api/src/importers/alc/datasource.js
+++ b/api/src/importers/alc/datasource.js
@@ -1,0 +1,166 @@
+const defaultLog = require('../../utils/logger')('alc-csv-datasource');
+const RECORD_TYPE = require('../../utils/constants/record-type-enum');
+
+class AlcCsvDataSource {
+  /**
+   * Creates an instance of DataSource.
+   *
+   * @param {*} taskAuditRecord audit record hook for this import instance
+   * @param {*} auth_payload information about the user account that started this update
+   * @param {*} recordType record type to create from the csv file
+   * @param {*} csvRows array of csv row objects to import
+   * @memberof AlcCsvDataSource
+   */
+  constructor(taskAuditRecord, auth_payload, recordType, csvRows) {
+    this.taskAuditRecord = taskAuditRecord;
+    this.auth_payload = auth_payload;
+    this.recordType = recordType;
+
+    // Only process records that are Status = 'Complete', Reason != 'File Closure', Record Type = Inspection
+    if (csvRows) {
+      this.csvRows = csvRows.filter(row => {
+        const reportStatus = row['status'];
+        const reason = row['reason'];
+        const type = row['record type'];
+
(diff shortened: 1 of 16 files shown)


marklise (changes requested):
I notice there's lat/lon in the CSV, but the import isn't setting those.

BcGovNeal:
> I notice there's lat/lon in the CSV, but the import isn't setting those.

Jira ticket says not to use those for privacy reasons.

marklise:
> > I notice there's lat/lon in the CSV, but the import isn't setting those.
> 
> Jira ticket says not to use those for privacy reasons.

Ah gotcha - we may want to address this in another ticket (ie: protecting centroids..)

marklise (approved):

mtCarto (approved):

api/src/importers/alc/datasource.js
```diff
@@ -0,0 +1,166 @@
...
+   * @param {*} csvRows array of csv row objects to import
+   * @memberof AlcCsvDataSource
+   */
+  constructor(taskAuditRecord, auth_payload, recordType, csvRows) {
+    this.taskAuditRecord = taskAuditRecord;
+    this.auth_payload = auth_payload;
+    this.recordType = recordType;
+
+    // Only process records that are Status = 'Complete', Reason != 'File Closure', Record Type = Inspection
+    if (csvRows) {
```
MaxWardle, line 20: Oh I like this in the constructor, so we can still do batch promises.

api/src/importers/alc/inspections-utils.js
```diff
@@ -0,0 +1,82 @@
...
+   * @memberof Inspections
+   */
+  transformRecord(csvRow) {
+    if (!csvRow) {
+      throw Error('transformRecord - required csvRow must be non-null.');
+    }
+
+    const inspection = { ...super.transformRecord(csvRow) };
+
+    inspection['_sourceRefStringId'] = Number(csvRow['record id']) || '';
```
MaxWardle, line 36: Shouldn't this be a string based on the name?

BcGovNeal: Yup I'll set it to a string to match the field name

MaxWardle (approved):
Looks crispy clean 👍

PR #782, opened 2021-02-19
increase max file upload size to 400mb

mtCarto:
[NRPT-692](https://bcmines.atlassian.net/browse/NRPT-692)

Increase doc upload limit to 400mb

Can create a ~350mb and 500mb file with
`dd if=/dev/zero of=./foo.pdf bs=3m count=128`
`dd if=/dev/zero of=./fooBig.pdf bs=4m count=128`

### DIFF
--- a/angular/projects/common/src/app/document-link-staging/document-link-staging.component.html
+++ b/angular/projects/common/src/app/document-link-staging/document-link-staging.component.html
@@ -18,7 +18,7 @@
             <label>Upload Document <span class="helper-text">(The name of the document is what will show publicly, please ensure your document is named appropriately prior to uploading.)</span></label>
             <app-file-upload
               [maxFiles]="params.maxFiles ? params.maxFiles : 5"
-              [maxSize]="params.maxSize ? params.maxSize : 25"
+              [maxSize]="params.maxSize ? params.maxSize : 400"
               [showInfo]="params.showInfo ? params.showInfo : false"
               [showList]="params.showList ? params.showList : false"
               [files]="recordFiles"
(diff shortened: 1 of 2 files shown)


BcGovNeal:
These 2 lines in the API seem like they could affect the upload size as well.  Although I didn't find them affecting anything locally.

https://github.com/mtCarto/NRPTI/blob/b709032e8436c9724164cb093b15faac2e41170c/api/app.js#L25-L26

mtCarto:
Good catch, I didn't notice anything locally either with the current settings. Increased it to match the new limit to make it consistent at least.

BcGovNeal (approved):

api/app.js
```diff
@@ -22,8 +22,8 @@ const DB_USERNAME = process.env.MONGODB_USERNAME || '';
 const DB_PASSWORD = process.env.MONGODB_PASSWORD || '';
 
 // Increase post body sizing
-app.use(bodyParser.json({ limit: '25mb', extended: true }));
-app.use(bodyParser.urlencoded({ limit: '25mb', extended: true }));
+app.use(bodyParser.json({ limit: '400mb', extended: true }));
+app.use(bodyParser.urlencoded({ limit: '400mb', extended: true }));
```
marklise, line 26: These entries are not quite the right spot to adjust file upload limits as they only restrict the size of the json or urlencoded based payloads.  FormData file uploads are handled separately from the bodyParser module.  This is why you're not seeing an effect when adjusting these values.    There are technically two spots to limit, 1) in the front end, which your earlier change covers, and 2) in the API where we actually receive the data.  We could use something like npm package `multer` in the api.  Mind you the `multer-s3` package might a be better option given what we are doing but I don't think it supports limiting filesizes.

marklise (changes requested):

marklise (approved):

PR #786, opened 2021-02-24
NRPT-666: LNG minimap added

cameronpettit:


mtCarto (approved):
Mini-map looks great, everything is working as expected. Only improvemnt would be the one UX issue you mentioned in the ticket, to hide the vector layers while flying.

marklise (approved):

marklise (approved):
Looks good, land when tests pass

PR #789, opened 2021-02-24
NRPT-295: Added ERA CSV importer

MaxWardle:
[NRPT-295](https://bcmines.atlassian.net/browse/NRPT-295)

Still working out unit tests


api/src/importers/era/tickets-utils.js
```diff
@@ -0,0 +1,100 @@
...
+        type: MiscConstants.IssuedToEntityTypes.Company,
+        companyName: csvRow['fc_client_name'] || ''
+      };
+    }
+
+    if (entityType === MiscConstants.IssuedToEntityTypes.Individual) {
+      ticket['issuedTo'] = {
+        type: MiscConstants.IssuedToEntityTypes.Individual,
+        dateOfBirth: null,
+        fullName: csvRow['fc_client_name'] || '',
```
BcGovNeal, line 89: I don't think setting `fullName` here works because it gets overwritten by the other name values further down the chain.  You can verify this by looking at the records that are Individual types.  Might be easier to just assign `fc_client_name` to `firstName`

api/src/importers/era/base-record-utils.js
```diff
@@ -0,0 +1,157 @@
...
+  transformRecord(csvRow) {
+    if (!csvRow) {
+      throw Error('transformRecord - required csvRow must be non-null.');
+    }
+
+    return {
+      _schemaName: this.recordType._schemaName,
+      recordType: this.recordType.displayName,
+
+      sourceSystemRef: 'era-csv'
```
BcGovNeal, line 51: Not sure if it's a big deal, bu the ticket says source should be `nro-era-csv`

api/src/importers/era/tickets-utils.js
MaxWardle: ooh good catch, I'll update

api/src/importers/era/base-record-utils.js
MaxWardle: was just following the pattern of the other importers, I can change it though

BcGovNeal: It's probably okay to just leave it

MaxWardle: okok

PR #796, opened 2021-03-04
NRPT-693 Added LNG link to NRCED records

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-683

### DIFF
--- a/angular/projects/public-nrpti/src/app/pipes/project-link.pipe.ts
+++ b/angular/projects/public-nrpti/src/app/pipes/project-link.pipe.ts
@@ -1,0 +1,27 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+/**
+ * Replace project names with links to LNG or CGL sites
+ *
+ * @export
+ * @class LinkifyPipe
+ * @implements {PipeTransform}
+ */
+@Pipe({ name: 'projectLink' })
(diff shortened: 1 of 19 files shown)


mtCarto (approved):

angular/projects/public-nrpti/src/app/pipes/project-link.pipe.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!str) {
+      return '';
+    }
+
+    if (str === 'LNG Canada') {
+      return `<a href="https://lng.gov.bc.ca/project/1/overview" target="_blank">${str}</a>`;
+    }
+
+    if (str === 'Coastal Gaslink') {
+      return `<a href="https://lng.gov.bc.ca/project/2/overview" target="_blank">${str}</a>`;
```
marklise, line 22: Can we make these relative to the current host URL?

BcGovNeal: I thought about that, but we don't have dev and test LNG URLs.  So there's no good way to generalized it in the NRCED app unless we put it into the config service.  I don't think it's worth the extra code for a minor feature like this.  The only users affected by this would be dev and test users.

marklise (approved):

PR #797, opened 2021-03-05
NRPT-619: implement admin component for LNG map updates

mtCarto:
- [NRPT-619](https://bcmines.atlassian.net/browse/NRPT-619)

### DIFF
--- a/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
+++ b/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
@@ -51,12 +51,6 @@ export class CommunicationsComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
       this.commPackage = res.communicationsPackage.COMMUNICATIONS;
-      // if (res && res.lngMapData && res.lngMapData[0] && res.lngMapData[0].data) {
-      //   this.lngMapInfo = res.lngMapData[0].data.searchResults;
-      // } else {
-      //   alert('Error: could not load LNG Map data');
-      // }
-
       this.selectedApplication = this.route.snapshot.params.application.toUpperCase();
       this.buildForm();
     });
--- a/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
+++ b/angular/projects/admin-nrpti/src/app/communications/communications.component.ts
@@ -51,11 +51,11 @@ export class CommunicationsComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
       this.commPackage = res.communicationsPackage.COMMUNICATIONS;
-      if (res && res.lngMapData && res.lngMapData[0] && res.lngMapData[0].data) {
-        this.lngMapInfo = res.lngMapData[0].data.searchResults;
-      } else {
-        alert('Error: could not load LNG Map data');
-      }
+      // if (res && res.lngMapData && res.lngMapData[0] && res.lngMapData[0].data) {
+      //   this.lngMapInfo = res.lngMapData[0].data.searchResults;
+      // } else {
+      //   alert('Error: could not load LNG Map data');
+      // }
 
       this.selectedApplication = this.route.snapshot.params.application.toUpperCase();
       this.buildForm();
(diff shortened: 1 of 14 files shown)


angular/projects/admin-nrpti/src/app/communications/lng-map-info/lng-map-info.component.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  async submit() {
+    this.loadingScreenService.setLoadingState(true, 'main');
+
+    if (
+      !this.mapForm.get('location').value ||
+      !this.mapForm.get('length').value ||
+      !this.mapForm.get('recentUpdates').value
+    ) {
+      alert('Please ensure your updates to the Section info are complete');
```
BcGovNeal, line 102: Should the code return here?  Currently it'll alert, but continue to save the empty values anyways.

How does the user "cancel" an update if `recentUpdates` can't be empty?

angular/projects/admin-nrpti/src/app/communications/lng-map-info/lng-map-info.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+      this.loadingScreenService.setLoadingState(false, 'main');
+    }
+
+    this.router.navigate(['communications', 'LNG']);
+  }
+
+  async cancel() {
+    // reset form values
+    this.buildForm();
```
BcGovNeal, line 125: `currentSection` values aren't updated after a form submit.  This is causing the reset to use values from before the form changes.

To reproduce:

1. Select a section
2. Update the description text and save
3. Hit Cancel
4. Now the description is reset back to the value in Step 1.  The expected value should be value from Step 2

angular/projects/admin-nrpti/src/app/communications/lng-map-info/lng-map-info.component.ts
mtCarto: ahh yes it should, I'll update this

angular/projects/admin-nrpti/src/app/communications/lng-map-info/lng-map-info.component.ts
mtCarto: Save should be triggering the resolver so that the description is update. I'll investigate this

marklise: Save should not be triggering a resolver at all.  The message should popup and no angular router action should take place.

marklise:
When hitting Save/Publish Map Updates and one of the required fields is missing information, the alert message comes up telling you to fill them out, but then a resolver is hit.  The angular router should not be using the resolver in this scenario.  The alert message should popup and when dismissed go right back to the form.

marklise:
1. enter text in recent updates (leave length empty)
2.  hit save
3. get alert telling you to fill out forms
4. dismiss the dialog
5. hit cancel map updates
6. navigate away
7. come back, your 'cancelled' changes are still there.

This is also likely due to the resolver being used in the previous comment.

angular/projects/admin-nrpti/src/app/communications/lng-map-info/lng-map-info.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+  }
+
+  async submit() {
+    this.loadingScreenService.setLoadingState(true, 'main');
+
+    if (
+      !this.mapForm.get('location').value ||
+      !this.mapForm.get('length').value ||
+      !this.mapForm.get('recentUpdates').value
```
BcGovNeal, line 110: I'm not sure we should check `recentUpdates` value here.  Currently it doesn't allow user to set an empty value if they want to remove an update.

angular/projects/admin-nrpti/src/app/communications/communications.component.ts
```diff
@@ -46,8 +51,13 @@ export class CommunicationsComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.route.data.pipe(takeUntil(this.ngUnsubscribe)).subscribe((res: any) => {
       this.commPackage = res.communicationsPackage.COMMUNICATIONS;
-      this.selectedApplication = this.route.snapshot.params.application.toUpperCase();
+      // if (res && res.lngMapData && res.lngMapData[0] && res.lngMapData[0].data) {
```
BcGovNeal, line 54: Please remove

BcGovNeal (approved):
Looks good! Only issue left is #799 has the same migration as this PR. It just depends on who merges first. The later merge should have the migration removed.

PR #798, opened 2021-03-08
NRPT-13 Add NRCED CSV Export

BcGovNeal:
https://bcmines.atlassian.net/browse/NRPT-13

- Added CSV export to NRCED.  Export button only shows if the results are being filter by keywords or the advanced filters.
- Added `compression` module to the API to enable gzip compression.  API responses for the CSV exports are around 6-8MB per request.  The `compression` module reduces these requests down to 500-700KB.

### DIFF
--- a/angular/projects/public-nrpti/src/app/records/utils/record-utils.ts
+++ b/angular/projects/public-nrpti/src/app/records/utils/record-utils.ts
@@ -215,8 +215,8 @@ function escapeCsvString(csvField: any): string {
 }
 
 /**
- * Escape special characters in the CSV fields so the final output can be opened
- * correcly in Excel.
+ * Create a file download in the browser and save the file to the user's
+ * default download directory
  *
  * @static
  * @param {string} filename CSV filename to be saved as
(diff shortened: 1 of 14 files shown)


angular/projects/public-nrpti/src/app/records/utils/record-utils.ts
```diff
@@ -43,4 +45,188 @@ export class RecordUtils {
...
+  // Escape commas
+  if (str.indexOf(',') > -1) {
+    str = `"${str}"`;
+  }
+
+  return str;
+}
+
+/**
+ * Escape special characters in the CSV fields so the final output can be opened
```
mtCarto, line 218: Update doc description for download

mtCarto (approved):
Exports look correct, functions as expected.

angular/projects/public-nrpti/src/app/records/utils/record-utils.ts
BcGovNeal: Fixed in https://github.com/bcgov/NRPTI/pull/798/commits/c3308888b1560b725839ee6c3de47c46c2c87768

marklise (approved):
👍

PR #809, opened 2021-03-18
NRPT-697: added check for roles when publishing names

MaxWardle:
To test switch your role to admin:nrced or admin:lng only and create a record with issued to type individual, age over 19 and publish to NRCED or LNG. Verify that the created records have public in the base read, but not in the issued to read.

To test the migration, you can use:

`db.nrpti.find({ issuingAgency: {$in: ["BC Oil and Gas Commission","Ministry of Energy, Mines and Low Carbon Innovation","Natural Resource Officers","Ministry of Forests, Lands, and Natural Resource Operations","Agricultural Land Commission","Ministry of Agriculture","BC Wildfire Service"]}, "issuedTo.type": "Individual","issuedTo.read": { $all: ["public"] }} ).pretty()`

and then there should be none after you run the migration.

### DIFF
--- a/api/src/controllers/post/administrative-penalty.js
+++ b/api/src/controllers/post/administrative-penalty.js
@@ -342,7 +342,7 @@ exports.createLNG = function(args, res, next, incomingObj) {
     administrativePenaltyLNG.publishedBy = args.swagger.params.auth_payload.displayName;
   }
 
-  administrativePenaltyLNG = BusinessLogicManager.applyBusinessLogicOnPost(administrativePenaltyLNG);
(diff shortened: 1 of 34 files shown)


api/src/controllers/post/administrative-penalty.js
```diff
@@ -13,7 +13,6 @@ const ADDITIONAL_ROLES = [
   utils.ApplicationRoles.ADMIN_AGRI,
   utils.ApplicationRoles.ADMIN_ENV_EPD,
   utils.ApplicationRoles.ADMIN_ENV_COS,
-  utils.ApplicationRoles.ADMIN_ENV_BCPARKS,
```
mtCarto, line 16: Is the parks role supposed to removed from this record type?

PR #811, opened 2021-03-22
NRPT-407: migration to add issuing agency and role to historical NRCED records

cameronpettit:


api/migrations/20210316214525-addIssuingAgencyAndRoleToHistoricalNRCED.js
```diff
@@ -0,0 +1,182 @@
...
+  const mClient = await db.connection.connect(db.connectionString, { native_parser: true });
+
+  try {
+    console.log('attempting migration...');
+
+    const nrpti = await mClient.collection('nrpti');
+
+    const ocersRecords = await nrpti.find({ sourceSystemRef: 'ocers-csv' }).toArray();
+
+    await ocersRecords.forEach(async record => {
```
BcGovNeal, line 47: The `await` here isn't actually waiting for the DB updates to finish.  One way around this is to convert the `forEach` to a `for...of` loop and await on the DB update code.

The `db.update` calls in `updateRecord` should have await, and `updateRecord` should be async.  Then each of the `updateRecord` calls should also have await.

I'd also use `updateOne` or `updateMany`.  `update` seems to be a deprecated function.

BcGovNeal (changes requested):

api/migrations/20210316214525-addIssuingAgencyAndRoleToHistoricalNRCED.js
```diff
@@ -0,0 +1,180 @@
...
+              write: role
+            }
+          }
+        );
+      } catch (error) {
+        console.log('Migration error: ', error);
+      }
+    } else {
+      try {
+        db.update(
```
BcGovNeal, line 159: Missed one :P

cameronpettit: oops.. I will get that

api/src/utils/constants/misc.js
```diff
@@ -18,6 +18,22 @@ const ApplicationRoles = {
 
 exports.ApplicationRoles = ApplicationRoles;
 
+const ApplicationAgencies = {
+  AGENCY_ALC: 'Agricultural Land Commission',
+  AGENCY_WF: 'BC Wildfire Service',
+  AGENCY_ENV_COS: 'Conservation Officer Service',
```
BcGovNeal, line 24: Use `Conservation Officer Service (COS)` to match other records in DB

api/src/utils/constants/misc.js
```diff
@@ -18,6 +18,22 @@ const ApplicationRoles = {
...
+  AGENCY_BCMI: 'Ministry of Energy, Mines and Low Carbon Initiatives',
```
BcGovNeal, line 25: Use `EMLI` to match other records in DB

BcGovNeal (approved):
Looks good, just couple minor (and obscure) issues left.

mtCarto (commented):
MIgration looks good, address the string changes suggested by Neal and this is good to go

mtCarto (approved):

PR #813, opened 2021-03-25
NRPT-697: added name redaction on issuingAgency and made datepicker years scrollable

MaxWardle:
[NRPT-697](https://bcmines.atlassian.net/browse/NRPT-697): this is a recreation of the last PR to cleanup the commits 

To test switch your role to admin:nrced or admin:lng only and create a record with issued to type individual, age over 19 and publish to NRCED or LNG. Verify that the created records have public in the base read, but not in the issued to read.

for the following issuing agencies you should see public in the issuedTo read array if the other age requirements are met:

`
exports.AUTHORIZED_PUBLISH_AGENCIES = [
  'BC Parks',
  'Climate Action Secretariat',
  'Conservation Officer Service (COS)',
  'EAO',
  'Environmental Protection Division'
];
`

Any edits should apply the correct roles to the issuedTo read array as well.

To test the migration, you can use:

`
db.nrpti.find({ issuingAgency: {$in: ["BC Oil and Gas Commission","Ministry of Energy, Mines and Low Carbon Innovation","Natural Resource Officers","Ministry of Forests, Lands, and Natural Resource Operations","Agricultural Land Commission","Ministry of Agriculture","BC Wildfire Service"]}, "issuedTo.type": "Individual","issuedTo.read": { $all: ["public"] }} ).pretty()
`

and then there should be none after you run the migration.

I also updated all of the datepickers to at least minDate vars. I checked with Kyle that dateIssued should not be able to be set in the future so added maxDate to those ones as well as the birthday datepickers.

Most of the datepickers are in admin nrpti, but there is one in public nrced

### DIFF
--- a/api/migrations/20210309170828-redactPublicNames.js
+++ b/api/migrations/20210309170828-redactPublicNames.js
@@ -36,15 +36,7 @@ exports.up = async function(db) {
       ...RecordTypeEnum.BCMI_SCHEMA_NAMES
     ];
 
-    const noPublishIssuingAgencies = [
-      'BC Wildfire Service',
-      'Ministry of Agriculture',
-      'Agricultural Land Commission',
-      'Ministry of Forests, Lands, and Natural Resource Operations',
-      'Natural Resource Officers',
-      'BC Oil and Gas Commission',
-      'Ministry of Energy, Mines and Low Carbon Innovation'
-    ]
+    const authorizedPublishAgencies = RecordTypeEnum.AUTHORIZED_PUBLISH_AGENCIES;
 
     const allRecords = await nrpti
     .find({
@@ -55,7 +47,7 @@ exports.up = async function(db) {
     const promises = allRecords.map(async (record) => {
 
       if ( record.issuingAgency
-        && noPublishIssuingAgencies.includes(record.issuingAgency)
+        && !authorizedPublishAgencies.includes(record.issuingAgency)
         && record.issuedTo
         && record.issuedTo.type === 'Individual'
         && record.issuedTo.read.includes('public') )
--- a/api/migrations/20210309170828-redactPublicNames.js
+++ b/api/migrations/20210309170828-redactPublicNames.js
@@ -1,0 +1,100 @@
+'use strict';
+
+let dbm;
+let type;
+let seed;
+
+const RecordTypeEnum = require('../src/utils/constants/misc');
+
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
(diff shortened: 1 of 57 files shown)


api/src/utils/put-utils.js
```diff
@@ -170,13 +170,39 @@ exports.editRecordWithFlavours = async function (args, res, next, incomingObj, e
...
+        // Grab the existing flavour object to get the issuingAgency and issuedTo if
+        // they don't exist in the flavourIncomingObj
+        const flavourRecordModel = mongoose.model(entry[0]);
+
+        const flavourExistingObject = await flavourRecordModel
+          .findOne({
+            _id: new ObjectId(flavourId)
+          });
+
+        const flavourExistingJSONObject = flavourExistingObject.toJSON();
```
BcGovNeal, line 185: Could just use `flavourExistingObject` itself without doing `.toJSON()`

api/migrations/20210309170828-redactPublicNames.js
```diff
@@ -0,0 +1,100 @@
...
+    const nrpti = await mClient.collection('nrpti');
+
+    const recordSchemas = [
+      ...RecordTypeEnum.MASTER_SCHEMA_NAMES,
+      ...RecordTypeEnum.LNG_SCHEMA_NAMES,
+      ...RecordTypeEnum.NRCED_SCHEMA_NAMES,
+      ...RecordTypeEnum.BCMI_SCHEMA_NAMES
+    ];
+
+    const noPublishIssuingAgencies = [
```
BcGovNeal, line 39: Maybe use the values from the `AUTHORIZED_PUBLISH_AGENCIES` constant instead of creating the inverse of it?  This avoids some issues such as inconsistent issuing agency names between API and UI (EMLI is stored as `EMLI` in the DB).  

https://github.com/bcgov/NRPTI/blob/master/api/src/utils/constants/misc.js#L164-L170

BcGovNeal (changes requested):
It works good  overall!  One issue I'm seeing is the roles that are connected to issuing agencies do not get added to the read and write arrays.

For example, create a `BC Wildfire Service` record as admin.  It will correctly not have `public`  in issuedTo read and write.  However it's missing `admin:wf` in the read and write, so a Wildfire user wouldn't be able to see the issueTo as well.

api/src/utils/put-utils.js
MaxWardle: I think this throws an error in processPutRequest in the record controller because the issuedTo isn't a json object, but like a mongoose object or something. Maybe there is a better way to handle this though?

api/migrations/20210309170828-redactPublicNames.js
MaxWardle: Good Idea, thanks!

BcGovNeal (approved):
Looks good!  As mentioned, the additional role logic will be handled in another ticket:  https://bcmines.atlassian.net/browse/NRPT-669

marklise (approved):

PR #829, opened 2021-04-13
NRPT-712 public get controllers now determine target subset based on schemaName

cameronpettit:


BcGovNeal (approved):

mtCarto (commented):
Search controller looks good, the redactedSubset is still getting populated with the schemas listed in the skip set. Is the intention to keep them out of the subset altogether?

cameronpettit:
I kind of assumed that whether or not we would be populating `redactedrSubset` would be addressed in Max's accompanying ticket, but I haven't discussed that with him yet. https://bcmines.atlassian.net/browse/NRPT-711?atlOrigin=eyJpIjoiZTUxNDBmMTM3Nzg4NDMwMzgzN2Q5YjgyNjAyM2MwNDQiLCJwIjoiaiJ9

mtCarto (approved):

PR #846, opened 2021-05-04
NRPT-757: moved legislation field on records to array

MaxWardle:
Do not merge until [NRPT-758](https://bcmines.atlassian.net/browse/NRPT-758) is complete, as this ticket will break all of the front ends with anything regarding a legislation field.

[NRPT-757](https://bcmines.atlassian.net/browse/NRPT-757): This ticket is to update the api and data on all records with a legislation field. We are preparing to accept multiple legislations per record and therefore are moving the legislation objects into an array.


api/migrations/20210427172703-legislationToArray.js
```diff
@@ -0,0 +1,135 @@
...
+    let numUpdatedRRS = 0;
+
+
+    const promisesNrpti = allRecordsNrpti.map(async (record) => {
+
+      if ( record.legislation ) {
+
+        let newLegislation = [record.legislation];
+
+        numUpdatedNrpti++;
```
marklise, line 55: Technically, this is only counting the number attempted to be updated.  You'll want to check the return value of `updateOne` below for `modifiedCount` and push that value into the records updated tracker.

api/migrations/20210427172703-legislationToArray.js
```diff
@@ -0,0 +1,135 @@
...
+    let numUpdatedNrpti = 0;
+    let numUpdatedRRS = 0;
+
+
+    const promisesNrpti = allRecordsNrpti.map(async (record) => {
+
+      if ( record.legislation ) {
+
+        let newLegislation = [record.legislation];
+
```
marklise, line 54: There's a bit too much whitespace rows here, no more than 1 space between code and I suggest this:

```const promisesNrpti = allRecordsNrpti.map(async (record) => {
      if ( record.legislation ) {
        let newLegislation = [record.legislation];
```

api/migrations/20210427172703-legislationToArray.js
```diff
@@ -0,0 +1,135 @@
...
+    console.log(`Finished transferring ${numUpdatedNrpti} records legislation fields in NRPTI `);
+
+
+    const promisesRRS = allRecordsRRS.map(async (record) => {
+
+      if ( record.legislation ) {
+
+        let newLegislation = [record.legislation];
+
```
marklise, line 90: Too much whitespace here

api/migrations/20200709183253-loadMemDocs.js
```diff
@@ -331,12 +332,12 @@ async function createMineDocument(nrpti, nrptiMine, collection, collectionDoc, n
...
-    Key: s3Key,
-    Body: minioObject.Body,
-    ACL: 'authenticated-read'
-  }).promise()
+  // await s3.upload({
+  //   Bucket: OBJ_STORE_BUCKET,
+  //   Key: s3Key,
+  //   Body: minioObject.Body,
+  //   ACL: 'authenticated-read'
+  // }).promise()
```
marklise, line 340: I don't think any of these changes should be part of this PR - it's an old migration.

MaxWardle: Good catch thanks

api/migrations/20210427172703-legislationToArray.js
MaxWardle: Thanks, fixed and will keep in mind for the future

api/migrations/20210427172703-legislationToArray.js
```diff
@@ -73,23 +53,28 @@ exports.up = async function(db) {
               legislationDescription: ""
             }
           }
-        );
+        ).then( result => {
+          if (result.modifiedCount > 0 ) {
+            numUpdatedNrpti++;
+          }
+        });
```
marklise, line 60: Don't mix async/await and .then.  Better to do something like:

```
const result = await redacted_record_subset.updateOne(...
if (result.modifiedCount > 0) {
  numUpdatedNrpti++;
}
...
```

Since the await "waits" for the execution to finish before moving onto the next line, the return value is the payload of the function call so you can just assign the return value into an object and check it on the next line.

MaxWardle: nice catch tyty

BcGovNeal (approved):
Looks good with the latest changes

marklise (approved):

PR #882, opened 2021-07-15
NRPT 361 amalgamation of NRPTI issuing agency values

cameronpettit:
A bit of a pervasive bug to fix - `issuingAgency` values in the db have been amalgamated such that there is one entry per agency. The search filters currently implemented do not allow searching for agency names with commas in them - they are parsed at the commas into individual queries, so the db agency names do not contain commas.

The codebase has been combed and a wrapper function is used everywhere the issuing agency is displayed. There are two wrapper functions: One to display the full agency name (with commas) and the other to display the agency acronym, the latter being used for BCMI and the former everywhere else. 

There is a migration to update the existing db entries and the importers & record add/edit pages have been updated to maintain 1 name per agency.

### DIFF
--- a/api/migrations/20210712174600-mergeIssuingAgencyNames.js
+++ b/api/migrations/20210712174600-mergeIssuingAgencyNames.js
@@ -77,6 +77,23 @@ exports.up = async function (db) {
         { $set: { issuingAgency: 'Ministry of Agriculture Food and Fisheries' } }
       );
 
+      // We have to change 'agency' fields for _schemaName = CollectionBCMI as well - 'agency' fields only exist within the CollectionBCMI schema.
+
+      await currentCollection.updateMany(
+        { agency: 'EMLI' },
+        { $set: { agency: 'Ministry of Energy Mines and Low Carbon Innovation' } }
+      );
+
+      await currentCollection.updateMany(
+        { agency: 'EAO' },
+        { $set: { agency: 'Environmental Assessment Office' } }
+      );
+
+      await currentCollection.updateMany(
+        { agency: 'ENV' },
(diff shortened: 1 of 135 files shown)


marklise (approved):

api/migrations/20210712174600-mergeIssuingAgencyNames.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
+
+// list all collections that require this migration
+const collections = ['nrpti']
```
BcGovNeal, line 4: Should include redacted_record_subset as well

BcGovNeal (changes requested):
I think the values in the collection agency picklist should be updated to be consistent with records. Will also need to include migrations for the collections. 
https://github.com/cameronpettit/NRPTI/blob/239b42696a9fd9741bdf4820be1dd1d9b0fca1b2/angular/projects/common/src/app/utils/record-constants.ts#L269

api/migrations/20210712174600-mergeIssuingAgencyNames.js
cameronpettit: Updated `collectionAgencyPicklist` and added `redacted_record_subset` to migration

BcGovNeal: The picklist change looks good!  However the `CollectionBCMI` schema uses `agency` field instead of `issuingAgency` so the migration script still needs to account for them

cameronpettit: Good catch! Should have this fixed now in the migration.

BcGovNeal (approved):
Looks good!

PR #968, opened 2022-06-10
Update READMEs

norrisng-bc:
The install instructions are mostly unchanged. However, prerequisite version numbers are now expressed as a range (instead of a minimum version), as newer versions will break the application. 

Links were also added/fixed as required.

`npm` was also rewritten to simply require the latest version, instead of a specific one - this allows for the use of `corepack`, which greatly simplifies installing Yarn.


norrisng-bc:
_Side note:_ rebased to sign previously-pushed commits

PR #975, opened 2022-07-07
Task/setup sustainment apps locally/972

dmitri-korin-bcps:
I had build and runtime errors when I attempted to set up NRPTI locally. See the attachments. These code changes are to fix the errors.
[Build Error.log](https://github.com/bcgov/NRPTI/files/9066679/Build.Error.log)
[Runtime Error.txt](https://github.com/bcgov/NRPTI/files/9066680/Runtime.Error.txt)

### DIFF
--- a/angular/package.json
+++ b/angular/package.json
@@ -133,6 +133,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "~1.19.1",
     "puppeteer": "^2.1.1",
+    "sass": "^1.53.0",
     "ts-node": "~7.0.0",
     "tsickle": "0.34.0",
     "tslint": "~5.20.1",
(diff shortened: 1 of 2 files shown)


angular/package.json
```diff
@@ -133,6 +133,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "~1.19.1",
     "puppeteer": "^2.1.1",
+    "sass": "^1.53.0",
```
norrisng-bc, line 136: The `package-lock.json` has this:
```
"node-sass": "4.13.1"
```
Not too familiar with frontend, but given that Dart-Sass (`sass`) is the successor to `node-sass`, wouldn't there be a potential to break the CSS, given that it's essentially bumping up the version number?

norrisng-bc:
@dmitri-korin-bcps Yarn was something the previous devs decided on, doesn't really matter to me though we'd probably have to update the READMEs (and make sure the updated instructions work) if we fully switch to one or the other.

I think I would add `node-sass` to`package.json` for now, fully switching might be better handled in a separate ticket/sprint.

PR #985, opened 2022-08-05
Add environment variables for NRIS-EMLI importer

norrisng-bc:
As mentioned previously, the NRIS-EMLI importer is ready but was left inactive, as NRIS-EMLI and BCMI are not yet ready.

The EMLI importer is affected by the following environment variables:

* [`angular/projects/admin-nrpti/src/env.js`](https://github.com/bcgov/NRPTI/blob/task/research-enabling-emli-importer/981/angular/projects/admin-nrpti/src/env.js):
  ```js
  window.__env.FEATURE_FLAG = {
    "nris-emli-importer": false
  };
  ```
  * No changes were made here.

* [`tools/api/api.dc.yaml`](https://github.com/bcgov/NRPTI/blob/task/research-enabling-emli-importer/981/tools/api/api.dc.yaml), for the `nrpti-api` container:
  ```yaml
  - name: NRIS_EMLI_DOCUMENT_BINARIES_ENABLED
    value: 'false'
  - name: NRIS_EMLI_API_ENDPOINT
    value: 'https://api.nrs.gov.bc.ca/nrisws-api/v1/emprInspections'
  ```
  * [The existing code references the above 2](https://github.com/bcgov/NRPTI/blob/task/research-enabling-emli-importer/981/api/src/integrations/nris-emli/datasource.js), but they weren't defined in the DeploymentConfig. This PR adds them.
  * If `NRIS_EMLI_DOCUMENT_BINARIES_ENABLED` isn't defined as an environment variable (as was the case previously), it defaults to `false`.
  * `NRIS_EMLI_API_ENDPOINT` isn't strictly needed per se (the URL's hardcoded as a fallback), but this allows for changing the API endpoint in the future (NRIS-EPD, an existing importer, does this as well).

**TL;DR:**
* This PR adds the environment variables needed to enable the NRIS-EMLI importer.
* Toggling `nris-emli-importer` and `NRIS_EMLI_DOCUMENT_BINARIES_ENABLED` should be sufficient to enable it.

### DIFF
--- a/tools/api/api.dc.yaml
+++ b/tools/api/api.dc.yaml
@@ -134,6 +134,11 @@ objects:
                   value: ${KEYCLOAK_REALM}
                 - name: KEYCLOAK_ENABLED
                   value: 'true'
+                # Toggle to enable NRIS-EMLI importer
+                - name: NRIS_EMLI_DOCUMENT_BINARIES_ENABLED
+                  value: 'false'
+                - name: NRIS_EMLI_API_ENDPOINT
+                  value: 'https://api.nrs.gov.bc.ca/nrisws-api/v1/emprInspections'
               envFrom:
                 - prefix: NRIS_
                   secretRef:


Christopher-walsh22 (approved):

dmitri-korin-bcps (commented):
Just curious if/how you tested it. Did changing environment variable make importer work?

norrisng-bc:
> Just curious if/how you tested it. Did changing environment variable make importer work?

To be honest I haven't tested it, since the data source isn't getting turned on till September. 

I couldn't test `nris-emli-importer` (i.e. the frontend) either, since it affects the "Imports" page, which doesn't seem to be available outside of Prod.

PR #1100, opened 2023-09-08
1073 nrced about page update

davidclaveau:
Changelog:
* Update references from BC Oil and Gas Commission to BC Energy Regulator 
* Update wording around BC Energy Regulator change
* Fix links to new website

### DIFF
--- a/angular/projects/public-lng/src/assets/data/project1/authorizations.json
+++ b/angular/projects/public-lng/src/assets/data/project1/authorizations.json
@@ -183,7 +183,7 @@
         "complianceDocumentSubtype": "General",
         "OGCLabel": "Approved Project Permits",
         "name": " PE108469 PERMIT EMA",
-        "url": " https://www.bcogc.ca/node/13813/download"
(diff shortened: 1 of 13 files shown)


angular/projects/public-lng/src/assets/data/project1/authorizations.json
Christopher-walsh22, line 1: Link to document 13006 does not work. Do we want to just remove this link?

Christopher-walsh22 (approved):

sanjaytkbabu (approved):

angular/projects/public-lng/src/assets/data/project1/authorizations.json
RFK250: @Christopher-walsh22 Noticed this ticket was blocked, presumably by the above comment - you've probably already figured this out, but just in case, that's just on the LNG public page. It's scheduled to be decommissioned I believe, so you probably shouldn't spend any time editing it. It was just those few instances on the NRCED About page that the ticket called for.

PR #1120, opened 2023-10-04
[NRPTI-1045][FEATURE] - issuing agency update

sanjaytkbabu:
## Pull Request Standards

- [x] The title of the PR is accurate
- [x] The title includes the type of change [`HOTFIX`, `FEATURE`, `etc`]  
- [x] The PR title includes the ticket number in format of `[NRPTI-###]`
- [ ] Documentation is updated to reflect change

# Description

- { issuingAgency is now stored in database and retrieved via api calls in nrpti (all pages), nrced(all pages) and nrpti importers }
- { Include any screenshots necessary }


LocalNewsTV (changes requested):
## Feedback

+ Love the new JSDoc Headers, major improvement in the codebase from before.

We should use a constants object with all the agency codes in it, currently if I were to update the 'AGENCY_FNLR_NRO" alone, I would have to change that value 20 times across 11 files. 

If we use a constants file for agency codes we can prevent issues that can occur from using a hard coded string in several places:
- Typos
- Reduce code complexity
- Make maintenance faster and easier
- Reduce potential bugs from future feature updates and the above

## Currently there is a bug in the new features!

- If I change the name of an agency it runs successfully ✅
- If I change the name of an agency a `second` time on the same page, no network request is sent, a false confirmation message appears, values do not change.
- If I enter an updated agency name, `do not` select an agency I want to update, I still receive a false positive from both the API and the frontend (Or just the frontend, if this is a subsequent request on the same page)

LocalNewsTV (approved):

PR #1234, opened 2024-06-12
[BUG][NRPTI-1231] Fix bug where manually added records are not appearing in the nrced csv export

sggerard:

* Added function in backend to get act code from act title using mapping table 'acts_regulations_mapping'
* Added logic in backend to check legislation acts when searching
* Removed function from frontend as not to cause duplication and references in tests

### DIFF
--- a/angular/projects/public-nrpti/src/app/records/records-list/records-list-resolver.ts
+++ b/angular/projects/public-nrpti/src/app/records/records-list/records-list-resolver.ts
@@ -25,10 +25,6 @@ export class RecordsListResolver implements Resolve<Observable<object>> {
       schemaList = params.activityType.split(',');
     }
 
-    if (params.act) {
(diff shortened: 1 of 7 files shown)


angular/projects/public-nrpti/src/app/records/records-list/records-list-resolver.ts
```diff
@@ -25,10 +25,6 @@ export class RecordsListResolver implements Resolve<Observable<object>> {
       schemaList = params.activityType.split(',');
     }
 
-    if (params.act) {
-      params.act = RecordUtils.appendActCodesToActNames(params.act, this.factoryService);
```
acatchpole, line 29: appendActCodesToActNames was a function written just for this. it is used nowhere else and could be removed as well to avoid bloating the code base.

acatchpole (approved):

PR #1247, opened 2024-06-26
[BUG][NRPTI-1165] token refresh for core importer

acatchpole:
# Description
The API access token granted when doing the core-documents import is only valid for 900 sec (15 minutes). While this is not normally noticeable when only importing a couple documents, an import of all documents can take over an hour to do. 

- Updated the core-documents importer to set an expiration time when the api access token is gotten.
- Added a check that gets a new token if the expiration time has passed 
- (unrelated) also deleted the migration updateBCMICollectionTypes from a previous PR that has now been deemed unnecessary. This migration was never applied to the production db.


tools/README.md
```diff
@@ -66,7 +66,7 @@ Moving the **App** from `dev` (`:latest`) to `test`
 ```sh
 oc project f00029-tools 
 oc tag nrpti:test nrpti:test-backup 
-oc tag nrpti:lastest nrpti:test 
+oc tag nrpti:latest nrpti:test 
 ```
```
acatchpole, line 70: unrelated typo fix

api/src/integrations/integration-utils.js
```diff
@@ -92,12 +92,12 @@ exports.getRecords = async function(url, options = undefined) {
       throw new Error('coreLogin - unable to log in to Core API.');
     }
 
-    return payload.access_token;
-  }
+    return payload;
+  };
```
acatchpole, line 96: changed this to return the whole object because it includes the ttl for the token

api/src/integrations/core/datasource.js
```diff
@@ -43,7 +43,8 @@ class CoreDataSource {
 
     try {
       // Get a new API access token.
-      this.client_token = await getCoreAccessToken(CORE_CLIENT_ID, CORE_CLIENT_SECRET, CORE_GRANT_TYPE);
+      const apiAccess = await getCoreAccessToken(CORE_CLIENT_ID, CORE_CLIENT_SECRET, CORE_GRANT_TYPE);
+      this.client_token = apiAccess.access_token;
```
acatchpole, line 48: refactored due to the change below

api/src/integrations/core-documents/datasource.js
```diff
@@ -269,12 +271,51 @@ class CoreDocumentsDataSource {
...
+   * checks if the given time has passed
+   *
+   * @param {int} expiryTime the epoch time we are checking for
+   * @returns {boolean} true if the provided time is in the past
+   *
+   * @memberof CoreDocumentsDataSource
+   */
+  isAPITokenExpired(expiryTime) {
+    return Date.now() >= expiryTime;
+  }
```
LocalNewsTV, line 318: As an edge case, would you want to artificially increase `Date.now()` to account for time for the API request to be made?

LocalNewsTV (commented):
There's a lot of leftover console logs in here, not sure if it was intended, but flagging it in case, more specific to ones like "Token updating..." immediately followed by "Token updated!"

api/migrations/20240619192233-updateBCMICollectionTypeField.js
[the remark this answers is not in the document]
acatchpole: yeah. this migration was never run on prod

api/src/integrations/core-documents/datasource.js
acatchpole: that's the TIME_BUFFER. we request a refresh 30 seconds before we actually need it to allow for any delays

api/src/integrations/core-documents/datasource.js
```diff
@@ -68,17 +68,17 @@ class CoreDocumentsDataSource {
...
-      this.status.itemTotal = permits.length;
+      const numPermits = permits.length;
+      this.status.itemTotal = numPermits;
       await this.taskAuditRecord.updateTaskRecord({ itemTotal: this.status.itemTotal });
 
       const permitUtils = new PermitUtils(this.auth_payload, RECORD_TYPE.Permit);
 
       // Push records to proccess into a Promise array to be processed in parallel.
-      for (let i = 0; i < permits.length; i += jobCount) {
+      for (let i = 0; i < numPermits; i += jobCount) {
```
fergmac, line 78: This could be a good opportunity to pull this code out into a function and write a test for it. This kind of code can be confusing to read. 

Not a blocker, I know we didn't write this code we are just updating the permit length reference.

PR #1248, opened 2024-06-27
[BUG][NRPTI-1242] Add support for creating & updating permits from multiple permit codes

sggerard:
## Pull Request Standards

# Description
Previously only the most recent valid permit was selected for a mine and records processed for the permit. However requirements changed and now both M and MX records need to be publicly available. A hotfix [hotfix](https://github.com/bcgov/NRPTI/pull/1162) was deployed but did not address the underlying problem leading to syncing issues. This PR attempts to properly implement a solution.

Replaces getMinePermit() with getMinePermits() and separates the logic out into multiple methods so that createMinePermit() and updateMinePermit() can process multiple mine permits instead of only the currently selected valid Mine Permit. This allows both MX and M permits and collections for a single mine to be processed during the cron run.

### DIFF
--- a/api/src/integrations/core/datasource.js
+++ b/api/src/integrations/core/datasource.js
@@ -250,6 +250,42 @@ class CoreDataSource {
   }
 
   /**
+   * Gets all valid permits for a Core mine. A valid permit must meet the following criteria:
+   *  - Must not be historical
+   * .
+   *
+   * @param {object} nrptiRecord NRPTI record
+   * @returns {object} Valid permit.
+   * @memberof CoreDataSource
+   */
+  async getMinePermits(nrptiRecord) {
+    if (!nrptiRecord) {
+      throw Error('getMinePermits - required nrptiRecord is null.');
+    }
+
+    // Get permits with detailed information.
+    const url = getIntegrationUrl(CORE_API_HOST, `/api/mines/${nrptiRecord._sourceRefId}/permits`);
+    const { records: permits } = await integrationUtils.getRecords(url, getAuthHeader(this.client_token));
+
+    return permits.filter(p => this.isValidPermit(p,nrptiRecord));
+  }
+
+  isValidPermit(permit,nrptiRecord){
+    //Mine must not be historical which is indicated by an authorized year of '9999' on the latest amendment.
+    if((permit.permit_amendments.length && !permit.permit_amendments[0].authorization_end_date) 
+      || permit.permit_status_code === 'O'){
+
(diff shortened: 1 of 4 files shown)


api/src/integrations/core/datasource.js
```diff
@@ -250,42 +250,65 @@ class CoreDataSource {
...
+    //Mine must not be historical which is indicated by an authorized year of '9999' on the latest amendment.
+    if((permit.permit_amendments.length && !permit.permit_amendments[0].authorization_end_date) 
+      || permit.permit_status_code === 'O'){
+
+      // Do not use 'G-4-352' for Lumby
+      // https://bcmines.atlassian.net/browse/NRPT-684
+      if (nrptiRecord.name === 'Lumby Mine' && permit.permit_no === 'G-4-352') {
+        return false;
+      }
+      return true;
```
LocalNewsTV, line 283: Since the if determines the return, couldn't the return just be the `!inverse` of the if.
If the intent isn't clear it could be assigned to a variable before the return

api/src/integrations/core/datasource.js
```diff
@@ -250,42 +250,65 @@ class CoreDataSource {
...
 
-    // First, any mines with 'X' as their second character are considered exploratory. Remove them.
-    // const nonExploratoryPermits = permits.filter(permit => permit.permit_no[1].toLowerCase() !== 'x');
+    return permits.filter(p => this.isValidPermit(p,nrptiRecord));
+  }
+
+  isValidPermit(permit,nrptiRecord){
+    //Mine must not be historical which is indicated by an authorized year of '9999' on the latest amendment.
+    if((permit.permit_amendments.length && !permit.permit_amendments[0].authorization_end_date) 
+      || permit.permit_status_code === 'O'){
```
LocalNewsTV, line 276: Just minor formatting inconsistencies to the rest of the file with regards to whitespace

LocalNewsTV (commented):
No red flags stand out :) Just some minor formatting inconsistency and one small suggestion

api/src/integrations/core/datasource.js
sggerard: It sure can! The sonarcloud robots agree with you too ;)

api/src/integrations/core/datasource.js
sggerard: Good spotting! I did see this when I reviewed the PR too, surprised the linter didn't catch it

acatchpole (approved):

PR #1251, opened 2024-07-04
[BUG][NRPTI-1250] Fix acts update function call

sggerard:

- Replaces call to this.parseTitleFromXML with exports.parseTitleFromXML as using 'this' is not valid outside of a class.
- This bug was preventing the cronjobs in the test environment from completing successfully


dallascrichmond (approved):
Looks good to me :)

Issue #1258, opened 2024-07-19
[FIX] Implement a better fix for the issuingAgency check in the business-logic-manager


acatchpole:
**Describe the task**
#1239 describes a bug where Administrative Penalty records were being improperly redacted. The check to see if the issuing agency was one that was allowed to publish non-redacted  records was faulty. It was often comparing the actual agency name to a list of agency codes. A function to get the agencyCode from the name was created. As it needed to look up values from the db, it was created as an async function. This created its own complications as it required many of the upstream functions to also be converted to async. Ultimately, the original refactoring effort was unsuccessful. While the records were no longer being improperly redacted, an error was being generated that made it so the record could not be saved. This error was not able to be addressed in a timely manner, so a bandaid fix was implemented. A synchronous version of the function was created that used hardcoded values instead of db lookup.

**Purpose**
The entire reason for agencyCodes is that sometimes agency names change or new agencies can be added. The bandaid function will eventually become out of date and some records will be improperly redacted.

**Acceptance Criteria**
- [ ] In the business-logic-manager, the check for if the issuingAgency is one with authority to publish names needs to be able to get also check the associated agencyCode.
- [ ] this must involve a call to the database as this is where up-to-date values are stored.

**Additional context**
- Below is the original async version of the getAgencyCodeFromName function. This can be used as a reference or implemented in its entirety if the upstream async/await issues can be resolved
```js
/**
 * Provides the agencyCode (an intermediate value) that is mapped to the given agency name
 * Eg: 'Agricultural Land Commission' should return 'AGENCY_ALC'
 * @param {string} agencyName - the name of an agency that provides records to NRPTI.
 * @returns {string} agencyCode - the intermediate code that maps to the agencyName
 */
exports.getAgencyCodeFromName = async function(agencyName) {
  const db = mongodb.connection.db(process.env.MONGODB_DATABASE || 'nrpti-dev');
  const collectionDB = db.collection('nrpti');

  try {
    let agencyList;
    // Obtain documents with Application Agency Schema
    let agencyDocuments = await collectionDB.find({ _schemaName: RECORD_TYPE.ApplicationAgency._schemaName }).toArray();
    // Using map function to iterate through the original array and creates
    // a new array with objects containing only the _id, agencyCode, and agencyName properties.
    agencyList = await agencyDocuments.map(item => ({
      agencyCode: item.agencyCode,
      agencyName: item.agencyName
    }));

    for (const agency of agencyList) {
[...]

PR #1259, opened 2024-07-19
[FIX][NRPTI-1239] Amp record redaction fix (temporary workaround)

acatchpole:
## Pull Request Standards

- [x] The title of the PR is accurate
- [x] The title includes the type of change [`HOTFIX`, `FEATURE`, `etc`]  
- [x] The PR title includes the ticket number in format of `[NRPTI-###]`
- [x] Documentation is updated to reflect change

# Description
When deciding if a record needs redaction, one of the checks is of the issuing agency and making sure that it has authority to publish records. BUT due to the change to using an intermediate code for representing agencies, this check was often comparing the actual agency name to the agency code and authorized agencies were not getting through. The first version of the fix included an async function that required many other upstream functions to be converted to async. Ultimately, this task was not able to be successfully completed in a timely manner. A bandaid function was created using hardcoded values. These will eventually be out of date and the bug behaviour will start appearing again. #1258 was created to address this tech debt.

- updated the check to see if the issuing agency has authority to publish by translating the agencyName to the associated agencyCode

### DIFF
--- a/api/src/controllers/agencies.js
+++ b/api/src/controllers/agencies.js
@@ -93,7 +93,16 @@ exports.protectedPut = async function(args, res, next) {
   next();
 };
 
-exports.getAgencyCodeFromName = function(agencyName){
+/** 
+ * gets the intermediate code (agencyCode) for the given agencyName
+ * A synchronous version of getAgencyCodeFromName. Instead of using the db to lookup values,
+ * this function uses some hardcoded constants.
+ * This function was implemented as a temporary fix because the async version proved difficult to implement.
+ * This should be replaced
+ * @param {*} agencyName 
+ * @returns {string} agencyCode
+ */
+exports.getAgencyCodeFromNameBandaid = function(agencyName){
   const AGENCY_NAME_CODE_MAP = [
     {
       agencyCode: 'AGENCY_ALC',
@@ -159,5 +168,5 @@ exports.getAgencyCodeFromName = function(agencyName){
     }
   }
 
-  throw new Error("Error in getAgencyCodeFromName : No code found for ", agencyName);
+  return agencyName; // if no matching code is found, just return the passed-in name
 };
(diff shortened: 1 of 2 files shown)


api/src/controllers/agencies.js
```diff
@@ -92,3 +92,81 @@ exports.protectedPut = async function(args, res, next) {
...
+      agencyCode: 'AGENCY_WLRS',
+      agencyName: 'Ministry of Water, Land and Resource Stewardship'
+    },
+    {
+      agencyCode: 'AGENCY_CAS',
+      agencyName: 'Climate Action Secretariat'
+    }
+  ];
+
+  for(const agency of AGENCY_NAME_CODE_MAP){
```
fergmac, line 165: It's a temporary fix so no need to change but because this is an array you could also use .find() array method:

`
const agency = AGENCY_NAME_CODE_MAP.find(agency => agency.agencyName === agencyName)
`

I would also assume that if no matching code is found we would return null?

api/src/utils/business-logic-manager.js
```diff
@@ -124,7 +125,8 @@ function isIssuedToConsideredAnonymous(issuedTo, issuingAgency) {
   }
 
   // check if the issuingAgency has legislative authority to publish names
-  if (issuingAgency && !constants.AUTHORIZED_PUBLISH_AGENCIES.includes(issuingAgency)) {
+  if (issuingAgency && !constants.AUTHORIZED_PUBLISH_AGENCIES.includes(issuingAgency)
```
fergmac, line 128: This is getting to be a lot to parse. We have a comment which is goo, we could also bind this to a constant:

`
hasLegislativeAuthorityToPublish = (
issuingAgency && 
!constants.AUTHORIZED_PUBLISH_AGENCIES.includes(issuingAgency) && !constants.AUTHORIZED_PUBLISH_AGENCIES.includes(agenciesController.getAgencyCodeFromNameBandaid(issuingAgency))
)
`

fergmac (dismissed):

fergmac (approved):

